PR #148, opened 2023-11-28
fix: allow user to override external role name for AWS cross-account access

randalljohnson:
Allow a user to specify which `runtime-manager-threeport-$name` role can assume the created role via IAM cross-account authentication.

### DIFF
--- a/cmd/tptctl/cmd/config_aws_cloud_account.go
+++ b/cmd/tptctl/cmd/config_aws_cloud_account.go
@@ -29,6 +29,7 @@ var providerRegion string
 var roleName string
 var awsAccountId string
 var defaultAccount bool
+var externalRoleName string
 
 // ConfigCurrentInstanceCmd represents the current-instance command
 var ConfigAwsCloudAccountCmd = &cobra.Command{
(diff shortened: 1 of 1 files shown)


cmd/tptctl/cmd/config_aws_cloud_account.go
```diff
@@ -126,11 +132,6 @@ var ConfigAwsCloudAccountCmd = &cobra.Command{
 			if err != nil {
 				cli.Error("failed to delete aws account", err)
 			}
-
-			err = provider.DeleteResourceManagerRole(roleName, *awsConf)
-			if err != nil {
-				cli.Error("failed to delete threeport role", err)
-			}
```
randalljohnson, line 133: This delete operation didn't need to be here, as the role wouldn't exist in the event that this code is executed.

cmd/tptctl/cmd/config_aws_cloud_account.go
```diff
@@ -113,9 +119,9 @@ var ConfigAwsCloudAccountCmd = &cobra.Command{
 			),
 			roleName,
 			awsAccountId,
-			provider.GetResourceManagerRoleName(requestedControlPlane),
+			externalRoleName,
 			*createdAwsAccount.ExternalId,
-			false,
+			true,
```
randalljohnson, line 124: This boolean enables the attachment of the `assume-any-role` policy, which allows anyone in an account to assume the created role by default.

PR #149, opened 2023-12-01
fix: add missing subdomain logic

randalljohnson:


cmd/tptctl/cmd/config_aws_cloud_account.go
```diff
@@ -118,10 +118,11 @@ var ConfigAwsCloudAccountCmd = &cobra.Command{
 				map[string]string{},
 			),
 			roleName,
+			*callerIdentity.Account,
```
randalljohnson, line 121: This refers to the `externalAccountId` parameter, which is needed to specify which external account we're allowing access to.

internal/gateway/domain_name_instance.go
```diff
@@ -67,130 +67,9 @@ func domainNameInstanceDeleted(
...
-	_, err = client.UpdateWorkloadInstance(r.APIClient, r.APIServer, workloadInstance)
-	if err != nil {
-		if errors.Is(err, client.ErrorObjectNotFound) {
-			// workload instance has already been deleted
-			return 0, nil
-		}
-		log.Error(err, "failed to update workload instance")
-		return 0, nil
-	}
-
```
randalljohnson, line 134: `domainNameInstanceDeleted` doesn't need to clean anything up anymore

internal/gateway/domain_name_instance.go
```diff
@@ -67,130 +67,9 @@ func domainNameInstanceDeleted(
...
-	// mark domain name instance as reconciled
-	domainNameInstanceReconciled := true
-	domainNameInstance.Reconciled = &domainNameInstanceReconciled
-	_, err = client.UpdateDomainNameInstance(r.APIClient, r.APIServer, domainNameInstance)
-	if err != nil {
-		return fmt.Errorf("failed to update domain name instance: %w", err)
-	}
-
-	return nil
-}
```
randalljohnson, line 193: Previously, this reconciler would update a `VirtualService` object with the correct domain. While implementing the `SubDomain`field on the `Gateway` object, I realized that reconciler can simply refer to the domain from a `DomainNameDefinition` and construct a full domain (either subdomain + domain, or just domain) based off the `GatewayInstance` subdomain + related `DomainNameDefinition` domain. As a result, this logic can be removed from this reconciler.

internal/gateway/domain_name_instance.go
```diff
@@ -352,15 +231,29 @@ func confirmDnsControllerDeployed(
...
+		}
+		externalDnsWorkloadDefinitionId = existingWorkloadDef.ID
+	}
+
 	// create external dns workload instance
 	externalDnsWorkloadInstance := v0.WorkloadInstance{
-		Instance:                    v0.Instance{Name: &workloadDefName},
+		Instance: v0.Instance{
+			Name: util.StringPtr(fmt.Sprintf("%s-%s", workloadDefName, *kubernetesRuntimeInstance.Name)),
+		},
```
randalljohnson, line 254: Append `KubernetesRuntimeInstance` name to `external-dns` for each instance to avoid name collisions in the Threeport API.

lander2k2 (approved):

HarisChoudhary (approved):

PR #174, opened 2023-12-23
feat: improve validation of workload instances

lander2k2:
This change implements validation of workload instances that ensures
only one workload instance can be deployed to each K8s runtime instance
when not using threeport-managed namespaces, i.e. when the user includes
a K8s namespace in the workload definition YAMLDocument.

Signed-off-by: Rich Lander <lander2k2@protonmail.com>


pkg/config/v0/workload.go
```diff
@@ -170,6 +170,41 @@ func (wi *WorkloadInstanceValues) Create(apiClient *http.Client, apiEndpoint str
...
+			if err != nil {
+				return nil, fmt.Errorf("failed to get kubernetes runtime instance ID: %w", err)
+			}
+			runtimeNames = append(runtimeNames, *runtime.Name)
+		}
+		for _, rName := range runtimeNames {
+			// if the workload instance is using a cluster that already has an
+			// instance for this definition, return error
+			if rName == *kubernetesRuntimeInstance.Name {
+				return nil, errors.New("only one workload instance per cluster may be deployed when a Kubernetes namespace is included in the workload definition YAMLDocument")
```
lander2k2, line 203: only one workload instance per cluster may be deployed when a Kubernetes namespace is included in the workload definition YAMLDocument\nif you would like to deploy this workload to the same K8s runtime (and continue to manage namespaces) you will need a new workload definition

pkg/config/v0/workload.go
```diff
@@ -170,6 +170,41 @@ func (wi *WorkloadInstanceValues) Create(apiClient *http.Client, apiEndpoint str
 		return nil, err
 	}
 
+	// check to see if client is managing namespace
+	if !*workloadDefinition.ThreeportManagedNamespace {
```
lander2k2, line 174: First check for nil value and retry until it is no longer nil.

randalljohnson (approved):
🚢

PR #185, opened 2024-01-30
chore: update go dependencies

lander2k2:
Includes an update to main package for threeport agent for compatibility with latest version of controller-runtime.


cmd/agent/main.go
```diff
@@ -80,11 +81,12 @@ func main() {
...
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
 		Scheme:                 scheme,
-		MetricsBindAddress:     metricsAddr,
-		Port:                   9443,
 		HealthProbeBindAddress: probeAddr,
 		LeaderElection:         enableLeaderElection,
 		LeaderElectionID:       "563c770d.threeport.io",
+		Metrics: metricsserver.Options{
+			BindAddress: metricsAddr,
+		},
```
randalljohnson, line 89: Do we need to expose a port for this?

randalljohnson (approved):

cmd/agent/main.go
lander2k2: The port that we were explicitly including was hard-coded to the default (9443).  That `Port` field was removed from that `Options` object and nested into a different object.  Since we're not parameterizing that value the way we are with the metrics port, I just removed it as it changes no functionality and would just muddy the code up for no reason.

PR #191, opened 2024-02-07
feat: add secret controller

randalljohnson:


pkg/api/v0/meta.go
```diff
@@ -0,0 +1,19 @@
...
+
+	// The object ID of the base object.
+	ObjectID *uint `json:"ObjectID,omitempty" query:"objectid" gorm:"not null" validate:"required"`
+
+	// The object type of the attached object.
+	AttachedObjectType *string `json:"AttachedObjectType,omitempty" query:"attachedobjecttype" gorm:"not null" validate:"required"`
+
+	// The object ID of the attached object.
+	AttachedObjectID *uint `json:"AttachedObjectID,omitempty" query:"attachedobjectid" gorm:"not null" validate:"required"`
+}
```
randalljohnson, line 19: Moved to separate file to decouple from `Workload` objects, as they are now generalized and can be used between any pair of object types.

PR #197, opened 2024-02-29
feat: refactor sdk codegen and introduce create

HarisChoudhary:


cmd/rest-api/main.go
```diff
@@ -126,47 +123,13 @@ func main() {
 	defer nc.Close()
 
 	// jetstream context
-	js, err := nc.JetStream(nats.PublishAsyncMaxPending(256))
+	js, err := InitJetStream(nc)
 	if err != nil {
-		e.Logger.Fatalf("failed to create jetstream context: %v", err)
+		e.Logger.Fatalf("failed to initialize nats jet stream: %v", err)
```
HarisChoudhary, line 128: ensure failures cause rest api failure

HarisChoudhary: confirmed causes crash

PR #198, opened 2024-03-01
chore: remove unused forward proxy object

lander2k2:


lander2k2:
During live review, it was proposed we move the forward proxy objects from the API into a `retired` directory in `cmd/database-migrator/migrations/retired`.  This lead to an inability to build the database-migrator binary:

db-migrator-build

So that file will need to remain in the API package.  I have added a comment to `pkg/api/v0/forward_proxy.go` to explain why it is there.

randalljohnson (approved):

HarisChoudhary (approved):

PR #265, opened 2024-04-02
feat: update nats version

HarisChoudhary:


lander2k2:
@HarisChoudhary I see the library version is updated.  I thought your intent was to update the NATS version running in the control plane.  If so, the image versions in `pkg/threeport-installer/v0/dependencies.go` will need updating

lander2k2 (approved):

PR #319, opened 2024-07-15
feat: implement certificate auth to Threeport DB

lander2k2:
This PR implements secure authentication to CRDB for the Threeport API.  It uses certificate authentication described in the cockroachdb docs [here](https://www.cockroachlabs.com/docs/stable/authentication#using-digital-certificates-with-cockroachdb).  New certificates for a `root` user and `threeport` user are generated for each control plane deployment.  The `root` user is used by the `db-init` init container.  The `database-migrator` and `api-server` containers use the `threeport` user.

This PR also includes bug fixes for the zap logger overrides for the gorm logger in `pkg/api-server/v0/database/database_gen.go`.  This bug had been causing panics when gorm returned errors.


HarisChoudhary (approved):
👍

randalljohnson (approved):
🚢

PR #329, opened 2024-11-14
feat: enable extensions to Threeport with SDK

lander2k2:


pkg/api/v0/workload.go
```diff
@@ -59,19 +59,6 @@ type WorkloadInstance struct {
 	// All events generated for the workload instance that aren't related to a
 	// particular workload resource instance.
 	Events []*WorkloadEvent `json:"Events,omitempty" query:"events" validate:"optional"`
-
-	// The threeport objects that are deployed to support the workload instance.
-	AttachedObjectReferences []*AttachedObjectReference `json:"AttachedObjectReferences,omitempty" query:"attachedobjectreferences" validate:"optional,association"`
-}
-
-// AttachedObjectReference is a reference to an attached object.
-type AttachedObjectReference struct {
```
randalljohnson, line 68: Is this reintroduced elsewhere?

randalljohnson (approved):
Two minor typos + one question regarding `AttachedObjectReferences`. Otherwise, looks good. Discussed everything else on the call.

pkg/api/v0/workload.go
lander2k2: Yes, it now lives in `pkg/api/v0/attached_object.go`

PR #353, opened 2025-06-03
fix: set helm manifest namespaces in post renderer

randalljohnson:
This fix ensures the correct namespaces are set in all Helm resources before applying them to a given cluster.

### DIFF
--- a/internal/helm-workload/post_renderer.go
+++ b/internal/helm-workload/post_renderer.go
@@ -70,6 +70,11 @@ func (p *ThreeportPostRenderer) Run(renderedManifests *bytes.Buffer) (*bytes.Buf
 			return nil, fmt.Errorf("invalid Kubernetes manifest: missing Kind or apiVersion")
 		}
 
+		// Set the namespace - SetNamespace will only set it for namespaced resources
+		if p.HelmWorkloadInstance.ReleaseNamespace != nil {
+			kubeObject.SetNamespace(*p.HelmWorkloadInstance.ReleaseNamespace)
+		}
+
 		// set label metadata on object to signal threeport agent
 		kubeObject, err = kube.AddLabels(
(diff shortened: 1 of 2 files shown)


internal/helm-workload/v0_helm_workload_instance.go
```diff
@@ -405,6 +406,11 @@ func getHelmActionConfig(
 	settings := cli.New()
 	settings.RepositoryConfig = HelmRepoConfigFile
 
+	// ensure helm repo config directory exists
+	if err := os.MkdirAll(filepath.Dir(settings.RepositoryConfig), 0755); err != nil {
+		return nil, nil, nil, nil, fmt.Errorf("failed to create helm repo config directory: %w", err)
+	}
+
```
randalljohnson, line 413: I received errors about this intermittently without this.

internal/helm-workload/post_renderer.go
```diff
@@ -70,6 +70,11 @@ func (p *ThreeportPostRenderer) Run(renderedManifests *bytes.Buffer) (*bytes.Buf
 			return nil, fmt.Errorf("invalid Kubernetes manifest: missing Kind or apiVersion")
 		}
 
+		// Set the namespace - SetNamespace will only set it for namespaced resources
+		if p.HelmWorkloadInstance.ReleaseNamespace != nil {
+			kubeObject.SetNamespace(*p.HelmWorkloadInstance.ReleaseNamespace)
+		}
+
```
randalljohnson, line 77: We're also setting the namespace on the `install.Namespace` field in `v0HelmWorkloadInstanceCreated`, however I wasn't able to track down why it wasn't actually getting set. It wasn't set when this function gets called, so that's why I added this step.

lander2k2 (approved):

PR #362, opened 2025-07-03
feat: improve terminology for image namespaces vs repos

lander2k2:
There are two places that we've used the term "image repo" where we should have used "image namespace".  The former refers to a specific image's repo (without the tag), e.g. `ghcr.io/threeport/threeport-rest-api`. The latter refers to a namespace which contains multiple image repos, e.g. `ghcr.io/threeport` which is the correct term in these contexts.

* the flag `control-plane-image-repo` has been replaced with `control-plane-image-namespace
* the field `ImageRepo` in the SDK config has been replaced with `ImageNamespace`

Note: variable names were not updated, just the user-facing terms.


randalljohnson (approved):
👍

PR #363, opened 2025-07-07
fix: an empty threeport config is no longer created if non-existent

lander2k2:
Previously, if the threeport config did not exist, an empty threeport config would be created - and then an error about missing config data would be returned to the user.  This confusing behavior has been replaced by an error message if the threeport config doesn't exist.


randalljohnson (approved):
📦

PR #366, opened 2025-09-05
chore: remove API resources for AWS S3 and RDS services

lander2k2:
This PR removes the native Threeport API objects for AWS RDS and S3 resources.  There is far too much surface area in the cloud providers' manage service catalogues to cover in core Threeport.  These will be relegated to implementation through Terraform and/or by way of extension modules in future.  The only support for cloud provider resources going forward will be for runtime environments.


lander2k2:
Things *not* included in this PR:
* The AWS wordpress workload tutorial in the user docs has a `TODO` placeholder and will be re-done in a separate PR
* The SDK tutorial for a Wordpress module has a `TODO` placeholder and will be re-done in a separate PR
* The sample manifests in the `samples` dir are left unchanged for now and will be updated when the workload tutorial is re-done.

randalljohnson (approved):

PR #371, opened 2025-09-30
feat: add tptctl replace command, remove describe command

lander2k2:
The tptctl CLI now has the update capability for all objects. This PR adds the `tptctl replace` command for all objects so that updates are possible with `PUT` requests.  It removes the `tptctl describe` command in favor of a more useful `tptctl get` command that now leverages an improved config package to present the config package's `*Config` objects back to the user.  This is important since these are the objects needed as inputs to `create` and `replace`.  So, now the user can retrieve, make modifications and then input the same object types.


randalljohnson (approved):
🚢

PR #375, opened 2025-10-17
feat: add descriptions to registered module objects

lander2k2:
This commit also improves the handling of registered module API routes. Previously, custom routes with ID paramters - ending with '/:id' - were being registered which is not necessary or desired.


randalljohnson (approved):
🚢

PR #376, opened 2025-10-17
fix: repair helm workload creation

lander2k2:
Helm workload creation was failing for two reasons:
1. The validation was checking for a defined kubernetes runtime, however it is not required since the default is used if not named.
2. When the helm workload instance is created, the config object was not returning the associated definition which was expcected by the defined instance abstraction.


randalljohnson (approved):
🚢

PR #377, opened 2025-10-26
feat: improve error message handling from threeport API gateways

lander2k2:
Until now, the response processing in the client library has been tailored to the response format from the Threeport API.  When there is an intermediate proxy, such as an API gateway, if the gateway returns an error message in the response body, it was lost.  Now, the client library will surface those errors in the response body to the user.


randalljohnson (approved):
👍

PR #378, opened 2025-10-30
feat: implement cursor-based pagination

lander2k2:
This commit adds pagination in all relevant API handlers.  It uses materialized views in the database to snapshot a set of results when the first page is requested and then uses unique ID cursors for clients to fetch subsequent pages of results.  This ensures no records will be missed or duplicated as is possible with offset-based pagination.


randalljohnson (approved):
👍

PR #379, opened 2025-11-04
fix: eliminate panic when fetching module objects

lander2k2:
Not all module objects (API objects) have a controller association in the database.  The objects that do not require controller reconciliation don't have a foreign key to the module controller table.  That was previously causing a panic when fetching module objects with tptctl. This commit fixes that problem.


randalljohnson (approved):
👍

PR #380, opened 2025-11-04
fix: remove materialized view cleanup from extension module APIs

lander2k2:
The Threeport SDK adds code to the main package of the Threeport API to clean up materialized view tables every 60 seconds.  These materialized views are used for cursor-based pagination.  The clean up only needs to be done by the Threeport API server.  The SDK was also adding it to the API server of each extension module.  This commit fixes this.


randalljohnson (approved):
👍

PR #381, opened 2025-11-05
fix: return non-zero exit from tptctl when no defined instance exists

lander2k2:
There was a bug in the config package that caused an error to be thrown when 'tptctl get [defined instance]' was run. The error was due to checks that returned errors when no records were found.  This check has been removed so that an empty array is returned from the config package. This restores the earlier behavior and retains the non-zero exit when no records found pattern for tptctl.


randalljohnson (approved):
👍

PR #382, opened 2025-11-06
chore: remove VERSION column from tptctl tabular output

lander2k2:
The version column was intended to inform the user which version of an API object was being displayed.  It was an attempt to future-proof for when multiple versions exist.  However, it was just confusing.  We will tackle how to distinguish different API object versions when there are more versions to manage.


randalljohnson (approved):
🚢

PR #384, opened 2025-11-28
chore: remove redundant method on ThreeportConfig

lander2k2:
There were two methods with identical functionality on the ThreeportConfig object:

* GetEncryptionKey
* GetThreeportEncryptionKey

This commit removes the first method and retains the second.


randalljohnson (approved):
👍

PR #387, opened 2025-12-17
feat: add support for Google Cloud Platform

lander2k2:


cmd/terraform-controller/image/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # generated by 'threeport-sdk gen' but will not be regenerated - can optionally be edited
-FROM golang:1.22-alpine as builder
+FROM golang:1.24-alpine as builder
```
lander2k2, line 2: Check that codegen has also been updated to make this change

cmd/tptctl/cmd/gcp_get_output.go
```diff
@@ -0,0 +1,99 @@
...
+
+// outputGetv0GcpProvidersCmd produces the tabular output for the
+// `get gcp-providers` command.
+func outputGetv0GcpProvidersCmd(
+	gcpProviders *[]config_v0.GcpProviderConfig,
+) error {
+	writer := tabwriter.NewWriter(os.Stdout, 4, 4, 4, ' ', 0)
+	// TODO: add columns for each field that users should see
+	// TODO: available fields are defined in the GcpProviderValues object in pkg/config/v0/gcp_provider.go
+	fmt.Fprintln(writer, "VERSION\t NAME\t AGE")
```
lander2k2: Add fields for output on these functions

docs/dev/release-checklist.md
```diff
@@ -94,8 +94,13 @@ versions as needed for future releases.
...
 - [] Update `kindest/node` image version used for local control planes.  It is defined in two
   places in `internal/provider/kind.go`.
-- [] Update container image version to latest for CockroachDB and NATS in
+- [] Update container image version to the latest stable version for CockroachDB and NATS in
   installer.
+- [] Update Pulumi version in all Dockerfiles to match the version of the `github.com/pulumi/pulumi/sdk/v3`
+  dependency in `go.mod`.  The relevant Dockerfiles live in:
+  - `cmd/oci-controller/images`
+  - `cmd/gcp-controller/images`
+- [] Update Go version on alpine images used for all image builds in `cmd/[component]/images`.
```
lander2k2, line 103: Add note to update that version in codegen as well

docs/docs/install/install-threeport-gcp.md
```diff
@@ -0,0 +1,12 @@
+# Install Threeport on GCP
+
```
lander2k2, line 2: Need to finish...

internal/gcp/v0_gcp_gke_kubernetes_runtime_instance.go
```diff
@@ -0,0 +1,562 @@
...
+
+const gkeStaleAckDurationSeconds = 600 // 10 minutes - Pulumi operations can take longer
+
+// v0GcpGkeKubernetesRuntimeInstanceCreated performs reconciliation when a v0 GcpGkeKubernetesRuntimeInstance
+// has been created.
+func v0GcpGkeKubernetesRuntimeInstanceCreated(
+	r *controller.Reconciler,
+	gcpGkeKubernetesRuntimeInstance *v0.GcpGkeKubernetesRuntimeInstance,
+	log *logr.Logger,
+) (int64, error) {
```
lander2k2, line 19: Align implementation pattern with OCI: https://github.com/threeport/threeport/blob/0.7/internal/oci/v0_oci_oke_kubernetes_runtime_instance.go

cmd/terraform-controller/image/Dockerfile
lander2k2: SDK already generates according to go version currently being used.  However, they need to be manually updated later when the go version is updated.

docs/dev/release-checklist.md
lander2k2: No need.  It captures it from go.mod automatically

internal/gcp/v0_gcp_gke_kubernetes_runtime_instance.go
lander2k2: Done

internal/provider/gke.go
```diff
@@ -0,0 +1,911 @@
...
+		return fmt.Errorf("failed to create IAM service client: %w", err)
+	}
+
+	return i.configureWorkloadIdentityBinding(iamService)
+}
+
+// loadGCPConfig reads the GCP configuration from gcloud CLI config files
+// or environment variables and populates KubernetesRuntimeInfraGKE struct fields.
+// It follows a similar pattern to loadOCIConfig in OKE.
+func (i *KubernetesRuntimeInfraGKE) loadGCPConfig() error {
```
lander2k2, line 528: In this function, we're picking up env vars for things like GCP project ID.  However, the `tptctl up` command requires a flag to set this.  The implication is that the env var would never get used.  Behavior should be: 1) if flags aren't provided, use env vars.  2) if flags are provided, they take precedence over env var.  3) if neither are provided (flag or env var) we need to return an error because SOME value is needed.

internal/provider/gke.go
```diff
@@ -0,0 +1,911 @@
...
+	// THIRD: Fall back to browser-based OAuth flow (scenario 1 - CLI only)
+	// This only works for CLI usage (tptctl), not for controllers
+	fmt.Println("GCP credentials not found or expired. Initiating authentication...")
+
+	// perform the OAuth flow
+	if err := performGCPOAuthFlow(ctx); err != nil {
+		return fmt.Errorf("failed to authenticate with GCP: %w", err)
+	}
+
+	fmt.Println("GCP authentication successful!")
```
lander2k2: Use `cli` package for formatting CLI message output

internal/provider/pulumi.go
```diff
@@ -0,0 +1,17 @@
...
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+)
+
+// GetPulumiStateDir returns the directory where Pulumi state is stored for Threeport.
+func GetPulumiStateDir() (string, error) {
+	homeDir, err := os.UserHomeDir()
```
lander2k2, line 13: This duplicates functionality here: https://github.com/threeport/threeport/blob/0.7/internal/provider/pulumi_workspace.go#L489
Consolidate this.

pkg/cli/v0/genesis_control_plane.go
```diff
@@ -40,6 +41,10 @@ type GenesisControlPlaneCLIArgs struct {
 	AwsRegion             string
 	OciRegion             string
 	OciConfigProfile      string
+	AwsRoleArn            string
+	AwsSerialNumber       string
```
lander2k2: Dafuq are these doing here???

internal/provider/gke.go
lander2k2: Flags in `up` command requiring GCP flags (when `-p gke` is used) must be removed.

A new optional flag `--gcp-account` must be added that allows the user to provide their account email.

An inline prompt must be added when using `tptctl up -p gke`.  The prompt will display the user's default GCP account email (if set).  If not set, an error that requests an env var or flag will be returned.  When the user confirms the email they want to use, then `tptctl` can kick off the oauth flow that opens the browser.

pkg/kube/v0/kube.go
```diff
@@ -0,0 +1,3 @@
+package v0
+
+const KubernetesDefaultVersion = "1.34.2"
```
lander2k2, line 3: Is this duplicated anywhere else?  Is this **really** the first time we've decided to set the canonical version for kube?

pkg/util/v0/build.go
```diff
@@ -76,6 +76,7 @@ func BuildImage(
 		"buildx",
 		"build",
 		"--load",
+		"--no-cache",
```
lander2k2: Remove this

randalljohnson (approved):
🚢

PR #393, opened 2026-04-21
fix: emit working codegen for tptctl config path feature

randalljohnson:
Fixes two codegen bugs that made `Tptctl.ConfigPath: true` in `sdk-config.yaml` produce non-compiling output.

In `pkg/sdk/v0/gen/cmd/cli/command.go` the generator referenced an identifier that was never declared. It now walks through the `*Config` wrapper in scope (for example `exampleConfig.Example.ExampleConfigPath` instead of `example.ExampleConfigPath`).

In `pkg/sdk/v0/gen/pkg/config/config.go` the emitted `*Values` struct never included the `<Group>ConfigPath *string` field that those assignments targeted. It now conditionally appends that field at both emission sites when the flag is set.


lander2k2:
The Threeport code that gets fixed by the SDK fixes should be included in the PR.

PR #395, opened 2026-04-21
feat: drive release image arch from sdk config

randalljohnson:
Adds a `ReleaseArch` field to `sdk-config.yaml` so downstream modules can declare their release image architecture. The generated release mage target reads this field instead of hardcoding `amd64`, and validates the value against a whitelist of supported architectures.

### DIFF
--- a/pkg/sdk/v0/config.go
+++ b/pkg/sdk/v0/config.go
@@ -28,6 +28,11 @@ type SdkConfig struct {
 	// A repository for each module will be created in this namespace.
 	ImageNamespace string `yaml:"ImageNamespace"`
 
+	// The CPU architecture used for release image builds, e.g. `amd64` or
+	// `arm64`. Defaults to `amd64` when unset. Dev image builds always use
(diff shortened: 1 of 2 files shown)


lander2k2:
We should either support an array of architectures or just default to all the popular ones.  In the near future we'll definitely want to support amd64 **and** arm64 in core Threeport

pkg/sdk/v0/config.go
```diff
@@ -28,6 +28,11 @@ type SdkConfig struct {
 	// A repository for each module will be created in this namespace.
 	ImageNamespace string `yaml:"ImageNamespace"`
 
+	// The CPU architecture used for release image builds, e.g. `amd64` or
+	// `arm64`. Defaults to `amd64` when unset. Dev image builds always use
+	// the host architecture.
+	ReleaseArch string `yaml:"ReleaseArch"`
```
randalljohnson, line 34: convert to slice of supported release architectures

PR #397, opened 2026-04-21
fix: skip nil and non-string values in RedactEncryptedValues to avoid…

randalljohnson:


randalljohnson:
Replaced by a superset branch — see the follow-up PR opened from fix-encrypted-field-update-path which includes these commits plus the Model+Updates/Changed() baseline restoration.

PR #398, opened 2026-04-23
fix: harden encrypted field redact and decrypt handling

randalljohnson:
Tightens the redact/decrypt/update path for encrypted fields. `RedactEncryptedValues()` now skips nil and non-string values to avoid panics on slice and map fields. `DecryptValues()` gains slice support and a nil-pointer guard. The update path uses `gorm.Model(...).Select("*").Omit(...)` and a `RedactedValuePlaceholder` so handlers don't overwrite existing ciphertext on updates that didn't change the encrypted value.


lander2k2:
It looks as though the generated code for the API handlers is included.  That code is boilerplate (always generated) so is pretty straight forward.

However, the tptctl code fixes are not included (only the SDK fix).  The tptctl code is scaffolding (initial file generated but never re-generated).  So the tptctl code changes have to be inserted as though it was generated by the SDK.

pkg/api-server/v0/handlers/aws_gen.go
```diff
@@ -307,7 +307,7 @@ func (h Handler) UpdateAwsEksKubernetesRuntimeDefinition(c echo.Context) error {
 	}
 
 	// update object in database
-	if result := h.DB.Model(&existingAwsEksKubernetesRuntimeDefinition).Updates(updatedAwsEksKubernetesRuntimeDefinition); result.Error != nil {
+	if result := h.DB.Model(&existingAwsEksKubernetesRuntimeDefinition).Updates(&updatedAwsEksKubernetesRuntimeDefinition); result.Error != nil {
```
randalljohnson, line 310: todo

pkg/api-server/v0/handlers/aws_gen.go
```diff
@@ -375,7 +375,8 @@ func (h Handler) ReplaceAwsEksKubernetesRuntimeDefinition(c echo.Context) error
 
 	// persist provided data
 	updatedAwsEksKubernetesRuntimeDefinition.ID = existingAwsEksKubernetesRuntimeDefinition.ID
-	if result := h.DB.Session(&gorm.Session{FullSaveAssociations: false}).Omit("CreatedAt", "DeletedAt").Save(&updatedAwsEksKubernetesRuntimeDefinition); result.Error != nil {
+	updateModel := h.DB.Session(&gorm.Session{FullSaveAssociations: false}).Model(&existingAwsEksKubernetesRuntimeDefinition)
+	if result := updateModel.Select("*").Omit("CreatedAt", "DeletedAt").Updates(&updatedAwsEksKubernetesRuntimeDefinition); result.Error != nil {
```
randalljohnson: revert, supply secret value on replace. check for guard on user supplying config with redacted value

pkg/api/v0/aws_validate.go
```diff
@@ -71,6 +71,11 @@ func (a *AwsProvider) BeforeCreate(tx *gorm.DB) error {
 				return fmt.Errorf("failed to get string value for %s: %w", field.Name, err)
 			}
 
+			// caller round-tripped without decrypting; preserve existing DB ciphertext
+			if underlyingValue == encryption.RedactedValuePlaceholder {
+				continue
```
randalljohnson: for create, throw an error. 

for update, throw an error on redacted placeholder.

for both cases, return http error (https://github.com/threeport/threeport/blob/0.7/pkg/api/v0/kubernetes_runtime_validate.go#L51)

pkg/sdk/v0/gen/cmd/cli/command.go
```diff
@@ -1339,6 +1339,24 @@ func GenCliCommands(gen *gen.Generator, sdkConfig *sdk.SdkConfig) error {
...
+										Qual(
+											"github.com/threeport/threeport/pkg/cli/v0",
+											"Error",
+										).Call(
+											Lit("config contains redacted secret values; re-run 'tptctl get ... --decrypt-secrets' and edit the decrypted output before replacing"),
+											Nil(),
+										),
+										Qual("os", "Exit").Call(Lit(1)),
+									),
+									Line(),
```
randalljohnson: regen

pkg/api-server/v0/handlers/module_gen.go
```diff
@@ -371,7 +371,8 @@ func (h Handler) ReplaceModuleApi(c echo.Context) error {
 
 	// persist provided data
 	updatedModuleApi.ID = existingModuleApi.ID
-	if result := h.DB.Session(&gorm.Session{FullSaveAssociations: false}).Omit("CreatedAt", "DeletedAt").Save(&updatedModuleApi); result.Error != nil {
+	updateModel := h.DB.Session(&gorm.Session{FullSaveAssociations: false}).Model(&existingModuleApi)
+	if result := updateModel.Select("*").Omit("CreatedAt", "DeletedAt").Updates(&updatedModuleApi); result.Error != nil {
```
randalljohnson: make sure CreatedAt isn't overwritten

pkg/api-server/v0/handlers/actuator_gen.go
```diff
@@ -303,8 +303,15 @@ func (h Handler) UpdateProfile(c echo.Context) error {
...
-	if result := h.DB.Model(&existingProfile).Updates(updatedProfile); result.Error != nil {
+	if result := h.DB.Model(&existingProfile).Updates(&updatedProfile); result.Error != nil {
 		h.Logger.Error("handler error: error updating object", zap.Error(result.Error))
+		// check if this is a custom HTTP error with specific status code
+		var httpErr *util_v0.HttpError
+		if errors.As(result.Error, &httpErr) {
+			return apiserver_lib.ResponseStatusErr(
+				httpErr.GetStatusCode(), c, nil, result.Error, objectType,
+			)
+		}
```
randalljohnson, line 314: check for consistency on pointers being passed into gorm

pkg/api/v0/aws_validate.go
```diff
@@ -124,6 +135,16 @@ func (a *AwsProvider) BeforeUpdate(tx *gorm.DB) error {
...
+			// reject the redacted placeholder — clients should send a real
+			// value to change the field, or omit it to leave it unchanged
+			if underlyingValue == encryption.RedactedValuePlaceholder {
+				return util.NewBadRequestError(
+					fmt.Sprintf(
+						"field %s contains redacted placeholder; provide a real value or omit the field",
+						field.Name,
+					),
+				)
+			}
```
randalljohnson: add BeforeUpdate() and BeforeCreate() to gcp_validate.go, add this change

pkg/sdk/v0/gen/cmd/cli/command.go
randalljohnson: revert this change, requires many manual changes to scaffolding

PR #400, opened 2026-04-23
fix: sign OKE token requests with TenancyOCID, not CompartmentOCID

randalljohnson:
OCI API request signing requires the tenancy OCID, not the compartment OCID, as the "tenancy" component of the key identifier. Using the compartment OCID works only when the compartment happens to be the root compartment (which equals the tenancy), and silently fails signature verification otherwise. Fixes authentication failures when OKE clusters live in a non-root compartment.


randalljohnson:
Superseded by #396, which already includes the tenancy OCID fix alongside the token-refresh transport rewrite. Closing.

PR #404, opened 2026-04-26
fix: capture created module object id in module registration

randalljohnson:
CreateModuleObject() returns the persisted object with its DB-assigned ID. The
generator was discarding it via `_`, leaving the local variable at the zero
value. Downstream code in the same generated function reads that variable's ID
to write foreign-key references for related rows, so module registration
silently bound dependents to ID 0.

This captures the return value and assigns it back into the local so the rest
of the generated flow sees the populated object.


lander2k2:
We should include the corrected generated code with this PR.  That is to say, build the SDK with this fix it in it and then run code generation so that the resulting `pkg/api-server/v0/module_gen.go` is fixed, too.

lander2k2:
> We should include the corrected generated code with this PR. That is to say, build the SDK with this fix it in it and then run code generation so that the resulting `pkg/api-server/v0/module_gen.go` is fixed, too.

Upon further inspection, this SDK change only affects module code and does **not** affect any Threeport code.

PR #409, opened 2026-05-11
refactor: drop redundant json field-name from api types

randalljohnson:
Drop redundant `json:"<FieldName>"` field-name spelling from api type tags. Per Go's encoding/json docs, `json:",omitempty"` uses the Go field name as the JSON key by default, so `json:"FieldName,omitempty"` was just spelling that default out. This PR simplifies every validate-tagged field to `json:",omitempty"`, preserving omitempty (which matters for partial PATCH bodies from controllers) and letting Go default the name.

The codegen now requires `json:",omitempty"` on every `validate:"required"`, `validate:"optional"`, and `validate:"optional,association"` field, and forbids `query:"..."` tags. The `PayloadCheck()` null-on-required guard rejects PATCH bodies that explicitly null out required fields, which is what omitempty avoids in the first place. The optional and association cases follow the same rule for wire-format consistency: without omitempty, a controller building a partial PATCH from a fetched-and-modified struct carries `"Field": null` entries for every untouched pointer, all of which the server has to discard. A companion QueryBinder in `pkg/api-server/lib/v0/binder.go` derives query keys from the lowercased Go field name, so explicit query tags are redundant noise at best and a silent rename hazard at worst.

Establishes a tag-order convention: `json -> validate -> gorm -> encrypt -> relationship -> persist`. The `pkg/sdk/v0/util.Tag(pairs ...[2]string)` helper preserves caller-supplied order in codegen output, since jen's default `.Tag(map[string]string)` sorts alphabetically. Scaffolded code now lands in convention order automatically.

Generalizes `persist:"false"` to a field-agnostic mechanism. Previously hardcoded to `Secret.Data`, the api server's `beforeCreate` hook now nulls any field tagged `persist:"false"` before insert, and the generated reconciler skips its pre-reconcile reload when any field carries the tag. Codegen emits a `PersistFalseFields()` provider method per type.

Adds `gorm:"not null"` on the Events struct's required fields. New docs page at `docs/docs/sdk/struct-tags.md` walks through every tag key, the ordering convention, codegen behavior, and pairing rules.

Until Go's `encoding/json/v2` proposal (https://github.com/golang/go/issues/71497, experimental in 1.25 per https://go.dev/blog/jsonv2-exp) supports a global "omitempty by default" knob, per-field tagging is the only path.


pkg/api/v0/aws.go
```diff
@@ -9,35 +9,35 @@ type AwsProvider struct {
...
 	// The region to use for AWS managed services if not specified.
-	DefaultRegion *string `json:"DefaultRegion,omitempty" query:"defaultregion" gorm:"not null" validate:"required"`
+	DefaultRegion *string `json:",omitempty" gorm:"not null" validate:"required"`
 
 	// See pkg/api/v0/aws_validate.go for validation logic
 	// of the AccessKeyID and SecretAccessKey fields.
 
 	// The access key ID credentials for the AWS account.
-	AccessKeyID *string `json:"AccessKeyID,omitempty" validate:"optional" encrypt:"true"`
+	AccessKeyID *string `json:",omitempty" validate:"optional" encrypt:"true"`
```
randalljohnson, line 27: How should queries for these be handled?

pkg/cli/v0/config.go
```diff
@@ -31,80 +31,80 @@ const (
...
 
 // EKSProviderConfig is the set of provider config information needed to manage
 // EKS clusters on AWs.
 type EKSProviderConfig struct {
-	AwsConfigProfile string `yaml:"AWSConfigProfile"`
-	AwsRegion        string `yaml:"AWSRegion"`
-	AwsAccountID     string `yaml:"AWSAccountID"`
+	AwsConfigProfile string `json:"AWSConfigProfile,omitempty"`
+	AwsRegion        string `json:"AWSRegion,omitempty"`
+	AwsAccountID     string `json:"AWSAccountID,omitempty"`
```
randalljohnson, line 91: is this case-inconsistency intentional?

docs/docs/sdk/struct-tags.md
```diff
@@ -0,0 +1,224 @@
...
+
+```go
+SSHPassword *string `json:",omitempty" validate:"optional" encrypt:"true"`
+```
+
+### `relationship`
+
+[Attached Object Reference](../architecture/threeport-core.md) (AOR)
+modeling. The value drives lifecycle behavior on the base object.
+Values:
```
randalljohnson, line 80: link docs/concepts/attached-object-reference.md here

pkg/api-server/v0/docs/swagger.json
```diff
@@ -16617,7 +16617,7 @@
                     "type": "integer"
                 },
                 "KubernetesWorkloadInstanceID": {
-                    "description": "The kubernetes workload instance this domain name belongs to.",
+                    "description": "The workload instance this domain name belongs to.",
```
randalljohnson: k8s rename regression

pkg/api/lib/v0/tags.go
```diff
@@ -9,8 +9,14 @@ const (
...
 )
 
+// JsonOmitempty is the json-tag substring every `validate:"optional"`
+// field must carry so absent values are dropped on serialize rather
+// than emitting JSON null / zero values.
+const JsonOmitempty = "omitempty"
+
 // RelationshipTypeKey is the modifier name in a relationship tag value
 // (e.g. `relationship:"requires;type:KubernetesRuntimeInstance"`).
 const RelationshipTypeKey = "type"
```
randalljohnson, line 22: check which fields in here can live in the sdk package instead

pkg/api/v0/aws.go
randalljohnson: leave as-is

pkg/cli/v0/config.go
randalljohnson: leave as-is for now

pkg/config/v0/domain_name_instance.go
```diff
@@ -87,7 +87,7 @@ func (d *DomainNameInstanceConfig) Get(
 			}
 		}
 
-		// get kubernetes workload instance
+		// get workload instance
```
randalljohnson: rename regression

pkg/config/v0/domain_name_instance.go
```diff
@@ -134,10 +134,10 @@ func (d *DomainNameInstanceConfig) Create(
 		return nil, fmt.Errorf("failed to get kubernetes runtime instance: %w", err)
 	}
 
-	// get kubernetes workload instance
+	// get workload instance
```
randalljohnson: rename regression

pkg/config/v0/domain_name_instance.go
```diff
@@ -134,10 +134,10 @@ func (d *DomainNameInstanceConfig) Create(
...
 	workloadInstance, err := client_v0.GetKubernetesWorkloadInstanceByName(apiClient, apiEndpoint, *domainNameInstanceValues.KubernetesWorkloadInstance.Name)
 	if err != nil {
-		return nil, fmt.Errorf("failed to get kubernetes workload instance with name %s: %w", *domainNameInstanceValues.KubernetesWorkloadInstance.Name, err)
+		return nil, fmt.Errorf("failed to get workload instance with name %s: %w", *domainNameInstanceValues.KubernetesWorkloadInstance.Name, err)
```
randalljohnson: rename regression

pkg/config/v0/domain_name_instance.go
```diff
@@ -224,10 +224,10 @@ func (d *DomainNameInstanceConfig) Replace(
 		kubernetesRuntimeInstance = *kubernetesRuntimeInst
 	}
 
-	// get kubernetes workload instance for update
+	// get workload instance for update
```
randalljohnson: rename regression

pkg/config/v0/domain_name_instance.go
```diff
@@ -224,10 +224,10 @@ func (d *DomainNameInstanceConfig) Replace(
...
 	workloadInstance, err := client_v0.GetKubernetesWorkloadInstanceByName(apiClient, apiEndpoint, *domainNameInstanceValues.KubernetesWorkloadInstance.Name)
 	if err != nil {
-		return nil, fmt.Errorf("failed to get kubernetes workload instance with name %s: %w", *domainNameInstanceValues.KubernetesWorkloadInstance.Name, err)
+		return nil, fmt.Errorf("failed to get workload instance with name %s: %w", *domainNameInstanceValues.KubernetesWorkloadInstance.Name, err)
```
randalljohnson: rename regression

pkg/config/v0/gateway_instance.go
```diff
@@ -85,7 +85,7 @@ func (g *GatewayInstanceConfig) Get(
 			}
 		}
 
-		// get kubernetes workload instance
+		// get workload instance
```
randalljohnson: rename regression

pkg/config/v0/gateway_instance.go
```diff
@@ -132,10 +132,10 @@ func (g *GatewayInstanceConfig) Create(
 		return nil, fmt.Errorf("failed to get kubernetes runtime instance: %w", err)
 	}
 
-	// get kubernetes workload instance
+	// get workload instance
```
randalljohnson: rename regression

pkg/config/v0/gateway_instance.go
```diff
@@ -132,10 +132,10 @@ func (g *GatewayInstanceConfig) Create(
...
 	workloadInstance, err := client_v0.GetKubernetesWorkloadInstanceByName(apiClient, apiEndpoint, *gatewayInstanceValues.KubernetesWorkloadInstance.Name)
 	if err != nil {
-		return nil, fmt.Errorf("failed to get kubernetes workload instance with name %s: %w", *gatewayInstanceValues.KubernetesWorkloadInstance.Name, err)
+		return nil, fmt.Errorf("failed to get workload instance with name %s: %w", *gatewayInstanceValues.KubernetesWorkloadInstance.Name, err)
```
randalljohnson: rename regression

pkg/config/v0/gateway_instance.go
```diff
@@ -218,10 +218,10 @@ func (g *GatewayInstanceConfig) Replace(
 		)
 	}
 
-	// get kubernetes workload instance for update
+	// get workload instance for update
```
randalljohnson: rename regression

pkg/config/v0/gateway_instance.go
```diff
@@ -218,10 +218,10 @@ func (g *GatewayInstanceConfig) Replace(
...
 	workloadInstance, err := client_v0.GetKubernetesWorkloadInstanceByName(apiClient, apiEndpoint, *gatewayInstanceValues.KubernetesWorkloadInstance.Name)
 	if err != nil {
-		return nil, fmt.Errorf("failed to get kubernetes workload instance with name %s: %w", *gatewayInstanceValues.KubernetesWorkloadInstance.Name, err)
+		return nil, fmt.Errorf("failed to get workload instance with name %s: %w", *gatewayInstanceValues.KubernetesWorkloadInstance.Name, err)
```
randalljohnson: rename regression

pkg/sdk/v0/config.go
```diff
@@ -15,55 +15,55 @@ import (
...
-	ImageNamespace string `yaml:"ImageNamespace"`
+	ImageNamespace string
 
 	// Details to be displayed with the API swagger docs that are served by the
 	// API server.
-	ApiDocs ApiDocs `yaml:"ApiDocs"`
+	ApiDocs ApiDocs
 
 	// The configuration of API objects used in the module.
-	ApiObjectConfig `yaml:",inline"`
```
randalljohnson, line 36: what is inline used for? do we have other examples of yaml tags that don't match the field name, which are important otherwise?

pkg/sdk/v0/gen/generator.go
```diff
@@ -134,6 +145,14 @@ type ApiObjectGroup struct {
...
 	// that need to verify a field's Go type read it here.
 	FieldTypes map[string]map[string]string
+
+	// StructEmbeds records the anonymous embed type names per struct in
+	// this group. Keyed by the embedding struct's name; the value is the
+	// list of type names embedded anonymously (e.g.
+	// "KubernetesWorkloadInstance" -> ["Common", "Instance", "Reconciliation"]).
+	// The collision check uses this to flatten embedded fields into the
+	// effective-key set.
+	StructEmbeds map[string][]string
```
randalljohnson, line 154: let's make sure all these additions are actually used

pkg/sdk/v0/gen/generator.go
```diff
@@ -935,6 +1082,102 @@ func (g *Generator) ValidateTags() error {
...
+						"%s and %s: effective query key %q collides",
+						a, b, key,
+					))
+					return
+				}
+				effective[key] = owner
+				if isOuter {
+					outerKeys[key] = fieldName
+				}
+			}
```
randalljohnson: let's reject query tags altogether, and drop this collision check logic too

pkg/sdk/v0/config.go
randalljohnson: `sigs.k8s.io/yaml` uses `json/encoding` internally, and its default behavior for embeds is `json:",inline"` when no tag is present. So we now get this behavior for free via the package import swap.

PR #414, opened 2026-05-27
feat: parallel image builds in mage AllImagesDev/Release

randalljohnson:
Adds parallel image build support to mage's AllImagesDev and AllImagesRelease via a new RunParallel util that mirrors tptdev build's worker-pool shape. Each worker pipelines build then push per component, so a fast image publishes before slower ones finish. CI now uses mage build:allImagesDev 8.


randalljohnson:
Superseded by #416, which lands the parallel-build and module-CLI work as two clean commits on top of 0.7.

PR #416, opened 2026-05-27
feat: parallel image builds and module install CLI alignment

randalljohnson:
Adds parallel image build support to mage's AllImages, AllImagesDev, and AllImagesRelease via a new RunParallel util in pkg/util/v0. Each worker builds then pushes its component, so fast images publish before slower ones finish. Worker count comes from the PARALLEL env var (default 1), so mage build:allImagesDev still runs sequentially and PARALLEL=4 mage build:allImagesDev fans out to four workers.

Replaces --dev with --debug on the module install command to align with tptdev. The flag sets pod imagePullPolicy to Always so a rollout re-pulls a moving dev tag. The module Installer struct gains a Debug bool field and a getImagePullPolicy() helper, and the four hardcoded IfNotPresent literals in the installer codegen route through it.


randalljohnson:
Closing — all five feature commits on this branch (74a98598, 97a871ce, ca4c8b52, 69e77c24, 9c23995f) are already ancestors of #421 (refactor-consolidate-dockerfiles), so the content lands when #421 merges. The three additional commits on this branch are 0.7 forward-merge plumbing and a generated-file regen, not new content. No separate merge needed.

PR #419, opened 2026-05-27
feat: unify mage install targets and add module plugin install

randalljohnson:
Routes all mage install targets through an `installDir()` helper that returns `$GOBIN` if set, otherwise `$GOPATH/bin` (via `go/build.Default.GOPATH`, which falls back to `~/go` when `$GOPATH` is unset).

`install:sdk`, `install:tptctl`, and `install:tptdev` now share one shape: a `Build.X` target compiles the binary to `bin/`, and `Install.X` copies it to the resolved install dir. `install:tptctl` previously required an explicit path argument and now takes no arguments; `install:tptdev` previously installed to `/usr/local/bin/tptdev` via `sudo cp` and now installs at the resolved install dir without sudo; `install:sdk` gains a `Build.Sdk` target and the same build-then-cp split.

To install somewhere other than the default, set `$GOBIN` for the invocation, for example:

```
GOBIN=~/bin mage install:tptctl
```

which installs the binary at `~/bin/tptctl`. `$GOBIN` is the standard Go knob for binary install location and applies to all three install targets.

For extension modules, the code generator now emits an `install:plugin` target that builds the module's tptctl plugin and installs it into the tptctl plugin directory, resolved from `$THREEPORT_PLUGIN_DIR` if set, otherwise `~/.threeport/plugins`. The generated target (`<module>` is the module's kebab-case name) is:

```go
// Plugin builds the tptctl plugin and installs it in the tptctl plugin directory.
func (Install) Plugin() error {
	build := Build{}
	if err := build.Plugin(); err != nil {
		return fmt.Errorf("failed to build tptctl plugin: %w", err)
	}

	pluginDir := os.Getenv("THREEPORT_PLUGIN_DIR")
	if pluginDir == "" {
		dir, err := cli.DefaultPluginDir()
		if err != nil {
			return fmt.Errorf("failed to determine tptctl plugin directory: %w", err)
		}
		pluginDir = dir
	}
	if err := os.MkdirAll(pluginDir, 0o755); err != nil {
		return fmt.Errorf("failed to create tptctl plugin directory: %w", err)
	}

	outputPath := filepath.Join(pluginDir, "<module>")
	installCmd := exec.Command(
		"cp",
		"bin/<module>",
		outputPath,
	)

	output, err := installCmd.CombinedOutput()
	if err != nil {
		return fmt.Errorf("install failed for tptctl plugin with output '%s': %w", output, err)
	}

	fmt.Printf("tptctl plugin installed and available at %s\n", outputPath)

	return nil
}
```


randalljohnson:
@greptileai please re-review

lander2k2 (approved):

PR #420, opened 2026-05-27
feat: add tptdev reinstall and --apis flag for scoped installs

randalljohnson:
Adds a `tptdev reinstall` command that picks up source changes on every installer-managed stateless resource in the control-plane namespace (selected via the `threeport.io/managed-by` label) by deleting each resource with a foreground cascade and recreating it through the install path. As a belt-and-suspenders measure, deployments are first scaled to zero so their pods exit before the delete fires, which keeps the teardown predictable when the API server is slow to honor the cascade. CRDB data, NATS data, the CA, and the `rest-api`'s external IP are deliberately left untouched.

Full delete + recreate is used rather than `strategy: Recreate` because the reinstall has to cover any spec change a dev's source edit can produce, including mutations to immutable deployment fields (selector labels, certain volume specs) that an in-place update would reject. `strategy: Recreate` only sequences pod replacement during an image swap; it does not let the controller adopt a changed selector or a restructured pod template. Tearing the resource down and reapplying it through the same install path the initial `tptdev up` uses sidesteps that class of failure entirely.

Also adds a `--apis` flag for scoped controller installs on `tptctl up`, `tptdev up`, and `tptdev reinstall`. The flag takes a comma-separated list of `sdk-config.yaml` api object group names (e.g. `kubernetes_workload,gateway`) and limits the install to those apis' controllers alongside the `rest-api` and `agent`. `tptdev reinstall` without `--apis` auto-detects the controller subset from the cluster's installer-managed deployments so it mirrors the current install rather than expanding it.

Controller selection is centralized in a new `pkg/threeport-installer/v0/selection.go` with `SelectControllersByGroup()`, `DetectInstalledControllerNames()`, and `SelectControllersForReinstall()`. The shared `ParseApis()` helper parses the comma-separated flag value once and is reused across the three `cmd` packages.


randalljohnson:
Moved to the fork: randalljohnson/threeport#24

PR #421, opened 2026-05-28
refactor: unify dockerfiles and move image builds to buildx CI

randalljohnson:
Consolidates 51 per-controller Dockerfile variants (Dockerfile, Dockerfile-alpine, Dockerfile-goreleaser per component) into a single multi-stage Dockerfile at the repo root. Each component picks its build target and entrypoint via build-args. Goreleaser drops its `dockers:` section; it's now scoped to tptctl and threeport-sdk archive builds. The SDK emits a release-only Dockerfile to module project roots via `GenRoot()` so module consumers get a working build context out of the box.

Container images build through docker buildx with native multi-arch (linux/amd64, linux/arm64). `util.BuildImage()` drives buildx with target, platforms, and build-args. The release stage stamps standard OCI image labels resolved from env vars or git rev-parse fallbacks so local and CI builds both produce labeled images.

Release CI splits into two reusable workflows. `_goreleaser.yml` runs the archive build for tptctl and sdk. `_build-images.yml` runs a two-arch matrix (linux/amd64, linux/arm64) in parallel, one runner per arch building all controllers via `mage build:allImages` and pushing single-arch tags as `<image>:<tag>-<arch>`. A follow-on `manifest` job runs `mage package:allManifests` to stitch the per-arch tags into `<image>:<tag>` via `docker buildx imagetools create`. CI knobs: `PARALLEL_GO_BUILD` caps `go build -p` workers; `PARALLEL_IMAGE_BUILD` caps concurrent buildx/manifest tasks.

PR #424, opened 2026-05-30
feat: add --apis flag for selective controller install

randalljohnson:
Adds an --apis flag to tptctl up, tptdev up, and tptdev reinstall that takes a comma-separated list of sdk-config api names (e.g. kubernetes_workload,gateway,kubernetes_runtime) and limits the install to those apis' controllers plus the rest-api and agent. The headline use case is tptctl up --control-plane-only --apis <list>, which installs threeport on a pre-existing kubernetes cluster with a chosen subset of controllers.

When --apis is omitted, behavior differs by command: tptdev reinstall auto-detects controllers currently deployed in the cluster (via the threeport.io/managed-by label) and re-applies only those; tptctl up and tptdev up default to the full ThreeportControllerList since there is no prior cluster state to detect from on a fresh install. The asymmetry is deliberate.

The selection logic lives in a new pkg/threeport-installer/v0/selection.go with three helpers. SelectControllersByGroup() is the pure filter shared by all three CLI consumers. DetectInstalledControllerNames() lists installer-managed Deployments to derive the reinstall default. SelectControllersForReinstall() ties them together. The pure helper is positioned for a future ControlPlaneInstance reconciler to call with the same semantics, so non-dev installs of threeport pick up the same mechanism without rework.

Stacked on PR #420. Base will retarget to 0.7 after #420 merges.

### DIFF
--- a/cmd/tptctl/cmd/up.go
+++ b/cmd/tptctl/cmd/up.go
@@ -18,9 +18,9 @@ import (
 	threeport "github.com/threeport/threeport/pkg/threeport-installer/v0"
 )
 
-// upGroupNames holds the --names flag value: a comma-separated list
+// upApis holds the --apis flag value: a comma-separated list
 // of sdk-config ApiObjectGroup names whose controllers to install.
-var upGroupNames string
+var upApis string
 
 // TODO: will become a variable once production-ready control plane instances are
 // available.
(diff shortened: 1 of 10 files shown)


randalljohnson:
Superseded by #426, which rolls up #420, #421, #424, and #425 into a single PR.

PR #425, opened 2026-05-31
feat: add mage imagesByApis targets and narrow integration test build

randalljohnson:
Adds a single source of truth for which threeport APIs the integration test exercises. A new mage build:imagesByApisDev <apis> target builds the rest-api, database-migrator, and one controller image per listed sdk-config api group, parsing a comma-separated list. mage test:up gains a matching <apis> argument that forwards to tptctl up --names so the deploy only installs those controllers. The integration tests workflow defines APIS once at the job level and references it from both the build and deploy steps.

A complementary multi-arch mage build:imagesByApis <apis> <repo> <tag> <arches> target is added alongside the dev variant so future workflows can build a configurable subset of images to any registry. It is not called by anything in this PR.

Stacked on PR #421. The util.BuildBinaries helper that the new generator code depends on lives there. Until #421 merges, this PR's diff includes that PR's content.


randalljohnson:
Superseded by #426, which rolls up #420, #421, #424, and #425 into a single PR.

PR #427, opened 2026-06-08
fix: correct certificate org for extension modules

sg40:
Previously, the organization added to the certificate for extension modules was the core API namespace - `threeport.io`. This change updates that to pull the API namespace for the extension module from the sdk config. Note: the `GenInstaller` function only generates code for extension modules, not core Threeport.


randalljohnson:
🚢

PR #429, opened 2026-06-09
refactor: install support services in control plane namespace

randalljohnson:
Collapses the `support-services-system` namespace into `threeport-control-plane`. The installer no longer creates a separate namespace for the `support-services-operator` and its resources; the operator pod, its service account, role bindings, CRDs, and namespaced manifests now install into the same namespace as everything else, and the teardown path only deletes the control plane namespace. The operator-managed component namespaces (`nukleros-gateway-system`, `nukleros-certs-system`, `nukleros-secrets-system`) are unchanged, as those are created by the operator from its component specs.


randalljohnson:
@greptileai review please

lander2k2 (approved):

PR #430, opened 2026-06-09
refactor: make mage build targets env-driven

randalljohnson:
Drops the `Dev` and `Release` suffixed variants of every mage build target and makes the surviving targets take no positional arguments. Build repo, tag, and architecture now come from the `IMAGE_REPO`, `IMAGE_TAG`, and `ARCH` environment variables. Each has a default: `IMAGE_REPO` falls back to the dev image namespace, `IMAGE_TAG` to the current version, and `ARCH` to the local machine architecture. A comma-separated `ARCH` such as `amd64,arm64` drives a multi-arch build, so a single `mage build:allImages` now covers the local dev, single-arch, and multi-arch cases.

This removes the `releaseArch` constant that hardcoded `amd64` for the release path. Builds default to the local architecture and only cross-compile when an architecture is passed, which lines up with the multi-arch buildx work now in 0.7.

Removed targets are the per-component `BinDev`, `BinRelease`, `ImageDev`, and `ImageRelease` variants plus `AllBinsDev`, `AllBinsRelease`, and `AllImagesDev`. The per-component `Bin` and `Image` targets, `AllBins`, `AllImages`, and the package manifest targets all remain, now env-driven. `DockerfileTarget` threading is unchanged, so the `release-pulumi` stage (oci, gcp) and the `release-terraform` stage (terraform) still build with their CLIs layered on.

Release CI passes `IMAGE_REPO`, `IMAGE_TAG`, and `ARCH` through the workflow env block instead of positional arguments. The test workflow's image build step becomes a bare `mage build:allImages`.

The source change is in the magefile generator under `pkg/sdk/v0/gen/root`; `magefile_gen.go` is regenerated output.


randalljohnson:
Moved to the fork: randalljohnson/threeport#23

PR #431, opened 2026-06-09
fix: reject foreign key changes on full replace

randalljohnson:
Clarifies how update hooks see object state and fixes a relationship foreign-key immutability hole that the unclear handling masked.

Adds a teaching comment above every `beforeUpdate` extension point in `pkg/api/v0/*_validate.go`, and the SDK generator emits the same block for newly scaffolded types, explaining that the hook receiver is the loaded DB row, the inbound values are in `tx.Statement.Dest`, and `tx.Statement.Changed` detects per-field changes. `lib.LoadUpdatedObjFromDB` is renamed to `lib.LoadObjFromDB` since it returns a clean snapshot of the current row whether called from a before-hook (pre-update) or an after-hook (post-update).

A new `lib.IncomingValues` helper names the values being written for an update: `tx.Statement.Dest` under a PATCH (where the receiver is the loaded committed row) or the receiver itself under a PUT (where `Model == Dest`). The encrypt and persist hooks now read through it instead of each re-deriving the redirect inline, and a single doc block in `pkg/api/lib/v0/update_hooks.go` documents the two GORM call shapes once.

The relationship FK immutability guard in `processRelationshipTaggedFieldsBeforeUpdate` relied on `tx.Statement.Changed`, which compares `Dest` against the model. Under a PUT (`Save`) the model is the inbound object, so `Changed` always reported no change and a previously-set relationship FK could be silently mutated or cleared through the `ReplaceX` endpoints. The guard now compares the committed pre-update values from `lib.LoadObjFromDB` against the inbound values from `lib.IncomingValues` directly, treating a nil inbound FK as an explicit clear only on a full replace (via `lib.IsFullReplace`) and as an absent, unchanged field on a patch. A Save-shape test covers the previously-untested PUT path alongside the existing Updates matrix.

### DIFF
--- a/pkg/api/lib/v0/encrypt_hooks.go
+++ b/pkg/api/lib/v0/encrypt_hooks.go
@@ -36,17 +36,11 @@ func encryptedFieldsFor(obj interface{}) []EncryptedField {
 	return p.EncryptedFields()
 }
 
-// ProcessEncryptTaggedFields encrypts fields tagged `encrypt:"true"`.
-// On update gorm fires the hook on Statement.Model (the loaded row);
-// redirect to Statement.Dest so the inbound plaintext is read instead
-// of stale loaded values. Create paths leave Model == Dest.
+// ProcessEncryptTaggedFields encrypts fields tagged `encrypt:"true"`,
+// reading the inbound values being written so it encrypts the caller's
+// plaintext rather than stale loaded values.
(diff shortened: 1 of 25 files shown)


randalljohnson:
Superseded by #435 after branch rename. The renamed branch and PR carry the same diff.

PR #432, opened 2026-06-09
test: cover persist-false hook

randalljohnson:
Adds `pkg/api/lib/v0/persist_hooks_test.go` covering `ProcessPersistFalseTaggedFields()` with the same in-memory sqlite + gorm pattern used by `encrypt_hooks_test.go`. Five scenarios pin the function contract: a type with a persist-false-tagged field drops the column on create while sibling fields persist, a type without tagged fields is a no-op, the column gets nulled regardless of whether the inbound value is nil or set, the current production update path leaves the tagged column intact because `ProcessCoreTaggedFieldsBeforeUpdate()` in `pkg/api/v0/tagged_fields.go` does not call the persist-false hook, and the dest-redirect branch already present in the function works end-to-end if the wiring is ever extended to `BeforeUpdate`. The update-path test docstring flags the current behavior as known-but-pending-design-decision so any change to the wiring surfaces in the test suite first.


randalljohnson:
Moved to the fork: randalljohnson/threeport#18

PR #433, opened 2026-06-09
feat: validate aws access-key pairing on update

randalljohnson:
Closes the gap where a `PATCH` on an `AwsProvider` could leave the row with only one of `AccessKeyID` / `SecretAccessKey` set, an invalid configuration that `beforeCreate()` already rejects. The new `beforeUpdate()` hook re-runs the same pairing check, but only when at least one of the two credential fields is in the patch. Unrelated updates keep working without having to resend credentials. For each key it takes the patch value when `tx.Statement.Changed()` reports a change and otherwise the loaded row value, then applies the same non-nil, non-empty set semantics as create. If exactly one resolved value is set, the update is rejected with the same `util.NewBadRequestError` message used at create time. Clearing both keys together remains allowed for accounts that switch to `RoleArn`-based auth.


randalljohnson:
Moved to the fork: randalljohnson/threeport#17

PR #435, opened 2026-06-10
fix: reject foreign key changes on full replace

randalljohnson:
Rejects relationship foreign-key changes on `PUT` (full replace). The previous immutability guard relied on `tx.Statement.Changed()`, which compares `Statement.Dest` against the model. Under `Save()` the model IS the inbound object, so the guard always reported no change and a previously-set relationship foreign key could be silently mutated or cleared through the `ReplaceX` endpoints.

The fix lands in two layers. The lower layer adds `lib.IncomingValues(tx)` (returns `tx.Statement.Dest`, which holds the patch struct under `PATCH` and the inbound object under `PUT`), `lib.IsFullReplace(tx)` (true when `Model == Dest`, i.e. `PUT`/`Save` shape), `lib.IsPartialUpdate(tx)` (true when `Model != Dest`, i.e. the `Updates` shape used by both `PATCH` and `DELETE`), and `lib.LoadObjectFromDB(tx)` (returns a clean snapshot of the current row via reflection on `tx.Statement.Model`, which is the loaded row under `PATCH` and the inbound object under `PUT`, both of which carry the ID). `LoadUpdatedObjFromDB` is renamed to `lib.LoadObjectFromDB` since it returns a clean snapshot regardless of hook timing. All four helpers take only `tx` and derive everything else from `tx.Statement`, so callers do not pass redundant context. The higher layer adds `lib.IsFieldChanged(tx, "FieldName") (bool, error)` for per-field change detection across both call shapes: under `PATCH` it validates the field name against the schema then routes to `tx.Statement.Changed()`; under `PUT` it loads the committed row and compares the named field by reflection. The helper returns an error if the field is unknown on either call shape, the row has no ID, or the DB load fails so callers can propagate the failure rather than silently allowing the update.

The `IsFullReplace` / `IsPartialUpdate` naming was a deliberate layer choice. The `PUT` / `PATCH` distinction lives at four different layers, and each layer carries a different naming option:

| layer             | partial detector  | on PATCH | on reconciled DELETE | caveat |
|-------------------|-------------------|----------|----------------------|--------|
| semantic (chosen) | `IsPartialUpdate` | true     | true                 | no     |
| GORM method       | `IsUpdates`       | true     | true                 | no     |
| threeport client  | `IsUpdate`        | true     | false                | yes    |
| HTTP verb         | `IsPatch`         | true     | false                | yes    |

The DELETE caveat applies to the HTTP-verb and threeport-client layers. A reconciled `DELETE`'s first call is a GORM `Updates` flag-write (`DeletionScheduled`, etc.) before the soft delete, so at the GORM and semantic layers it is genuinely partial (`Model != Dest`) and `IsUpdates` / `IsPartialUpdate` report true on it. Its HTTP method is `DELETE` (not `PATCH`) and its client call is `DeleteX` (not `UpdateX`), though, so `IsPatch(tx)` and `IsUpdate(tx)` return false and miss a call that is really partial; that miss is the caveat. The semantic layer describes the GORM call shape directly without leaking HTTP or threeport client vocabulary, and carries no `DELETE` caveat. `IsSave` / `IsUpdates` would also be caveat-free, but they leak GORM internals into the helper API which would tie callers to GORM rather than to a stable semantic contract.

The same `PUT`-blind bug was present in four other hooks that used `tx.Statement.Changed()` directly. Each becomes a check against `lib.IsFieldChanged()`: `KubernetesRuntimeDefinition.beforeUpdate()` guards `InfraProvider` and `HighAvailability`, `KubernetesRuntimeInstance.beforeUpdate()` guards `Location`, `AttachedObjectReference.beforeUpdate()` guards `Relationship`, and the `MachineWorkloadDefinition` / `MachineWorkloadInstance` env-validation gates skip when `Env` is unchanged. The `relationship_hooks.go` dispatcher keeps its inline comparison loop so it can diff every foreign key in one `LoadObjectFromDB()` read. The encrypt and persist hooks now read through `IncomingValues()` instead of each re-deriving the redirect inline.

A single doc block in `pkg/api/lib/v0/update_helpers.go` documents the two GORM call shapes once and includes the layer-choice rationale above. A teaching comment is emitted above every `beforeUpdate()` extension point so the reader sees, in context, which helpers are available for the method they are about to implement. Core scaffold:

```
// beforeUpdate runs before the KubernetesRuntimeInstance is updated.
//
// Receiver semantics depend on the GORM call shape; see
// pkg/api/lib/v0/update_helpers.go for the full model. The simplest
// per-field check is:
//   - lib.IsFieldChanged(tx, "FieldName"): works under both PATCH
//     and PUT, handles the DB load internally
// Lower-level helpers, useful when lib.IsFieldChanged doesn't fit:
//   - lib.IncomingValues(tx): values being written
//   - lib.IsFullReplace(tx): true on PUT (Save shape)
//   - lib.IsPartialUpdate(tx): true on PATCH/DELETE (Updates shape)
// Import:
//   lib "github.com/threeport/threeport/pkg/api/lib/v0"
func (k *KubernetesRuntimeInstance) beforeUpdate(tx *gorm.DB) error {
    // ...
}
```

The SDK scaffold is module-aware: it emits `lib.` in core threeport and `tpapi_lib.` in modules, matching the existing convention of aliasing threeport packages (e.g. `tpapi_v0`) from a module's perspective so the module's own `pkg/api/v0` package does not collide. Both variants include the `Import:` line since most `*_validate.go` files do not already have the lib package imported. Module scaffold:

```
// beforeUpdate runs before the ModuleObject is updated.
//
// Receiver semantics depend on the GORM call shape; see
// pkg/api/lib/v0/update_helpers.go for the full model. The simplest
// per-field check is:
//   - tpapi_lib.IsFieldChanged(tx, "FieldName"): works under both PATCH
//     and PUT, handles the DB load internally
// Lower-level helpers, useful when tpapi_lib.IsFieldChanged doesn't fit:
//   - tpapi_lib.IncomingValues(tx): values being written
//   - tpapi_lib.IsFullReplace(tx): true on PUT (Save shape)
//   - tpapi_lib.IsPartialUpdate(tx): true on PATCH/DELETE (Updates shape)
// Import:
//   tpapi_lib "github.com/threeport/threeport/pkg/api/lib/v0"
func (m *ModuleObject) beforeUpdate(tx *gorm.DB) error {
    return nil
}
```

New tests cover the helpers under both call shapes (diff cases, no-diff cases, clears, missing-field error path on both `PATCH` and `PUT`) and the previously-untested `PUT` paths on every migrated immutability check, including the `Location = nil` clear case that would have nil-derefed in the old error-message format.


randalljohnson:
@greptileai please review

randalljohnson:
@greptileai please review

randalljohnson:
@greptileai please review

pkg/api/lib/v0/object.go
```diff
@@ -56,25 +53,6 @@ func NewCleanSession(tx *gorm.DB) *gorm.DB {
...
-	// read the row by primary key into updatedObj
-	if err := NewCleanSession(tx).First(updatedObj, id).Error; err != nil {
-		return nil, fmt.Errorf(
-			"failed to reload %s/%d from database: %w",
-			util.ObjectTypeName(obj), id, err,
-		)
-	}
-	return updatedObj, nil
-}
-
```
randalljohnson, line 77: LoadUpdatedObjFromDB is moved to adjcent `update_helpers.go` as it's more aligned with those functions.

randalljohnson:
@greptileai one last review...

PR #439, opened 2026-06-11
docs: replace diagrams with draw.io on concepts/application-orchestration

andrey0718:


lander2k2 (changes requested):
Two problems that need to be fixed:
1. `GitOpsDelivery.drawio.svg` has a typo.  Step 4 text includes "custrom" instead of "custom."
2. `GitOpsDelivery.png` isn't deleted.

lander2k2 (approved):

PR #440, opened 2026-06-11
feat: add machine runtime infra fields and harden ssh path

randalljohnson:
This adds six infra provisioning fields to the machine runtime instance type: infra provider, region, machine type, image id, network id, and a resource inventory JSON column. The fields are plain data columns so a downstream branch can wire actual provisioning; imported machines leave them unset. Create validation requires a region whenever a provider is set, and update validation makes the five identity fields immutable once the row exists, including introducing them post-create. Swagger docs are regenerated; no other generated code changes because the generated reconciler and client operate on whole structs.

The SSH reconcile path is hardened alongside the new fields. Each reconcile pass runs under an injectable timeout context, connect and ping are wrapped so cancellation returns promptly without leaking goroutines or connections, the requeue delay is configurable for tests, and the deleted hook now records a warning event when a provider-provisioned machine is deleted before deprovisioning support lands, staying a no-op for imported machines.

Test coverage adds idempotency (no second host key PATCH on re-reconcile), ping failure, transport-level and HTTP 500 PATCH failures, event recording failure, context cancellation, connect and ping timeouts, deletion behavior, validate-hook tests including a resource inventory round-trip under sqlite, and two race-enabled concurrency tests that assert no data races, no connection leaks, and goroutine counts returning to baseline after a 200-wide burst.


randalljohnson:
@greptileai please review!

pkg/api/v0/machine_runtime.go
```diff
@@ -36,6 +48,16 @@ type MachineRuntimeInstance struct {
 	// connection. If not provided, captured on first connection.
 	HostKey *string `json:",omitempty" validate:"optional"`
 
+	// The provider region in which the machine is provisioned.
+	Region *string `json:",omitempty" validate:"optional"`
```
lander2k2, line 56: Consider using a `Location` field as with the [KRI](https://github.com/threeport/threeport/blob/0.7/pkg/api/v0/kubernetes_runtime.go#L49-L52) for consistency and provider abstraction.

PR #441, opened 2026-06-11
docs: replaced diagram with draw.io on concepts/dependencies

andrey0718:


lander2k2 (changes requested):
3 things that need to be change on the diagram:
1. "Front End Wordload" -> "Front End Workload". (typo)
2. Change to the original diagram.  Replace "Gloo Gateway" with just "Gateway"
3. Nitpick: Center the "Gateway" (was "Gloo Gateway") in the center of the box (it's currently close to the right edge).

lander2k2 (approved):

PR #442, opened 2026-06-11
docs: replace diagrams with draw.io on concepts/extensions page

sg40:


lander2k2:
There was a mistake on the original diagram `ThreeportExtensionCustomAWSServices.png`.  There was floating arrow that was supposed to connect Front End Workload -> Back End Workload.  Can you add that arrow.

lander2k2 (approved):

PR #443, opened 2026-06-12
test: cover oci oke lifecycle adapter

randalljohnson:
Adds unit tests for the OCI OKE lifecycle adapter in package oci, covering the InfraLifecycleProvider implementation on okeLifecycle, buildOkeInfra, and the two reconciler entry points. The 17 lifecycle methods are tested to their reachable boundary against an in-process httptest API stub and a fake JetStream context: the 14 API and NATS methods end to end, and the three that take an InfraProvider (OnCreateConfirmed(), SaveCreateOutputs(), OnDeleteConfirmed()) up to the local-failure boundary of their concrete OCI calls, since driving them further needs a real OCI network call and is left integration-only. The compartment-delete-error guard in OnDeleteConfirmed() is pinned by test so a future refactor cannot silently swallow it. Both notification publishers assert subject and payload.

No production code changes. Every config provider used in a test fails locally before any socket opens, so the suite is deterministic under go test -race -count=5 with no network I/O. mage build:tptctl and mage build:allBinsDev are green.

Two follow-ups belong to the shared internal/provider package and are out of scope here: a validity and overlap test for the unexported OKE CIDR constants, and bounding the OCI SDK calls that currently use context.Background() with no timeout. Both are noted for the core lifecycle test PR.

### DIFF
--- a/internal/oci/oke_lifecycle_test.go
+++ b/internal/oci/oke_lifecycle_test.go
@@ -1,0 +1,659 @@
+package oci
+
+import (
+	"errors"
+	"fmt"
+	"net/http"
+	"testing"
+	"time"
+
+	logr "github.com/go-logr/logr"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"gorm.io/datatypes"
+
+	notif "github.com/threeport/threeport/internal/oci/notif"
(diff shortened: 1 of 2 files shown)


randalljohnson:
Moved to the fork: randalljohnson/threeport#20

PR #444, opened 2026-06-12
test: harden infra provider lifecycle with tests and seams

randalljohnson:
This PR brings the shared internal/provider infrastructure lifecycle under unit and concurrency test, and adds a few small testability seams the tests need. It is additive and signature-preserving: InfraProvider keeps its zero-arg DeployInfra()/DestroyInfra(), and the handlers keep HandleInfraCreate(p, log)/HandleInfraDelete(p, log), so the existing gcp and oci adapter call sites compile unchanged and no other package is touched. That independence is the point: the test-hardening PRs in this series are meant to be reviewed in parallel without breaking a shared signature.

The seams. Lifecycle tuning (stale-ack threshold, refresh-ack interval, semaphore capacity, persist-failure retry count and delay) moves into a LifecycleConfig swapped in tests via setLifecycleConfig(); production keeps today's defaults exactly (semaphore 5, stale-ack 240s, refresh 60s, persist 30 times at 10s). A new exported NewPulumiWorkspace(name, project, ...opts) constructor with WithStateDirRoot() lets tests, and the future machine provider, point per-instance state at a temp dir instead of the real state root. An always-on atomic in-flight counter, read in tests via inFlightCount(), backs the scale assertion. A Clock seam (setLifecycleClock()) makes stale-ack timing deterministic in tests.

The headline scale test proves the invariant that matters for provisioning many machines at once: the semaphore acquire is non-blocking, so a full pool returns a requeue delay immediately and starts no goroutine. The test models the reconciler requeue loop driving 2000 instances and asserts peak concurrent executions never exceed the injected cap and that everything drains to baseline, clean under the race detector. It does not assert via runtime.NumGoroutine() and does not assume N callers yield N goroutines.

Two existing quirks are pinned with tests and flagged for discussion rather than changed here, since changing either is a behavior change that belongs in a coordinated follow-up: the create path returns an OnCreateConfirmed() error without persisting a create-failure, and the cross-replica delete guard has no CreationFailed term, so a freshly-acked-but-failed create blocks delete until the ack goes stale. One small guard is added: GetStateFilePath() now returns an error on an empty instance name instead of building a colliding path; this is unreachable in production, where the instance name is always set.

PR5 (feat-gce-machine-provider) consumes NewPulumiWorkspace()/WithStateDirRoot() and may branch from this branch while it is unmerged, so the branch is pushed early as a stable base.

### DIFF
--- a/internal/provider/execute_semaphore_test.go
+++ b/internal/provider/execute_semaphore_test.go
@@ -1,0 +1,419 @@
+package provider
+
+import (
+	"errors"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+	"gorm.io/datatypes"
+)
+
+// compile-time interface satisfaction check for the order recorder.
+var _ RefreshableProvider = (*orderRecordingInfra)(nil)
+
+// configureSemaphoreTest installs a lifecycle config with the given
+// semaphore capacity and fast persist retries, registers config
+// restoration, and registers a drain that runs before restoration so no
+// launch goroutine outlives the test's semaphore channel.
+func configureSemaphoreTest(t *testing.T, capacity int) {
+	t.Helper()
+	restore := setLifecycleConfig(LifecycleConfig{
+		StaleAckThreshold: 240 * time.Second,
+		RefreshInterval:   time.Hour,
+		SemaphoreCapacity: capacity,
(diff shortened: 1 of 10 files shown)


randalljohnson:
Moved to the fork: randalljohnson/threeport#21

PR #445, opened 2026-06-12
test: harden gke lifecycle adapter

randalljohnson:
This is PR3 of the infra-provider hardening series and is independent of the others; it mirrors the OKE adapter PR applied to the GCP GKE lifecycle adapter. It brings the two hand-written adapter files in internal/gcp under unit test by driving the real adapter against an in-process httptest stub of the threeport API plus a small fake infra provider, with no GCP, Pulumi, or NATS dependency.

Coverage spans all 17 InfraLifecycleProvider methods, the infra builder, the three reconciler entry points (created confirmed-noop, updated noop, deleted scheduled-and-confirmed noop, and deleted not-scheduled error), and an N-instance concurrency test under the race detector that proves the adapter holds no shared mutable state (distinct instances never interfere, and each PATCH carries its own ID).

Adapter-local bugs fixed: nil-pointer guards on required API fields so a malformed object returns a descriptive error instead of panicking the create goroutine; normalization of the create-completeness inventory check to trim whitespace and reject empty arrays and a quoted null literal, which is intentionally stricter than provider.inventoryCleared() (that one compares raw bytes), so the two deliberately diverge and may converge later; a comma-ok type assertion that returns an error instead of panicking on an unexpected infra provider type; and rejection of empty connection info on create confirmation. That last fix is a reconcile-semantics change, since the create state machine now errors and requeues where it previously succeeded with empty connection fields, and it is committed separately so it can be reviewed and reverted on its own.

Accepted coverage gaps: the GetConnection() call line in OnCreateConfirmed() is environment-dependent (real GCP auth and a cluster-manager client) and is covered indirectly through the extracted connection-update helper plus the lifecycle-provider hardening PR; the NotificationPayload() marshal-error branch is not forced because doing so would require an artificially invalid inventory. This PR makes no semaphore-cap, goroutine-leak, requeue, or stale-ack claim; those invariants live in internal/provider and belong to the lifecycle-provider hardening PR.

The GCE machine runtime PR later adds tests to this same Go package, so every package-level identifier in the new test file is gke-prefixed and no TestMain() is declared. That cross-PR guard is by convention, not enforced by the compiler, so the GCE PR uses the gce prefix and should re-run go test ./internal/gcp/... after forward-merging whatever has landed on 0.7, including this branch.

### DIFF
--- a/internal/gcp/gke_lifecycle.go
+++ b/internal/gcp/gke_lifecycle.go
@@ -2,6 +2,7 @@ package gcp
 
 import (
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -12,6 +13,7 @@ import (
 	v0 "github.com/threeport/threeport/pkg/api/v0"
 	client "github.com/threeport/threeport/pkg/client/v0"
 	controller "github.com/threeport/threeport/pkg/controller/v0"
+	kube "github.com/threeport/threeport/pkg/kube/v0"
 	notifications "github.com/threeport/threeport/pkg/notifications/v0"
 )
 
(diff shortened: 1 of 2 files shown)


randalljohnson:
Moved to the fork: randalljohnson/threeport#22

PR #446, opened 2026-06-12
docs: update diagram to be consistent with white backgrounds

sg40:


lander2k2:
This diagram was already an .svg and doesn't need to be replaced.

PR #449, opened 2026-06-13
fix: customize docker image for helm-workload-controller

lander2k2:
Fixes:
* Helm workload controller docker images was missing `/var/helm/.cache/helm` directory.  Also added `/var/helm/.local/share/helm` at Greptile suggestion.  Updated `BuildImage` mage target accordingly.
* `LoadImage` mage target not working for the custom dockerfile targets.
* A bug in helm-workload-controller and kubernetes-workload-controller that caused a failure managing the `ThreeportWorkload` K8s custom resource (used by threeport-agent) when a `HelmWorkloadInstance` or `KubernetesWorkloadInstance` is created, deleted, and then re-created.


randalljohnson (approved):
🚢

PR #451, opened 2026-06-15
docs: replace diagram with draw.io on aws/advanced-aws-setup page

sg40:


lander2k2 (changes requested):
Changes needed:
* The top grey box should include "AWS Account #1
* The account ID ("1111111111") is floating in space
* The text in the yellow IAM Role box is pushed too high and the text is wrapping at the end.   The box may need to be resized slightly.

lander2k2 (changes requested):
The top grey box says "AWS Account #2" but should say "AWS Account #1"

lander2k2 (approved):

PR #458, opened 2026-06-20
fix: correct helm storage namespace and RBAC for helm-workload-controller

lander2k2:
Helm release tracking secrets were being stored in the controller's default namespace rather than the release namespace, causing uninstalls to fail with ErrReleaseNotFound when the two differed. The storage namespace is now derived from ReleaseNamespace so Helm state is always co-located with the release resources.

The uninstall path no longer uses IgnoreNotFound=true; instead it explicitly checks for driver.ErrReleaseNotFound and logs a warning, making storage-namespace mismatches and other storage failures visible rather than masking them as successful uninstalls.

Installer changes add a shared ClusterRole for ThreeportWorkload (K8s CR used by threeport-agent) management and bind it to each controller's service account, with GKE Workload Identity support (kind:User subjects). The helm-workload-controller also receives a cluster-admin binding since it must manage arbitrary chart resources across any namespace.

### DIFF
--- a/internal/helm-workload/v0_helm_workload_instance.go
+++ b/internal/helm-workload/v0_helm_workload_instance.go
@@ -18,6 +18,7 @@ import (
 	"helm.sh/helm/v3/pkg/registry"
 	"helm.sh/helm/v3/pkg/releaseutil"
 	"helm.sh/helm/v3/pkg/repo"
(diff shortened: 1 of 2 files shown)


randalljohnson (approved):
🚢

Issue #462, opened 2026-07-01
Aliases for tptctl commands


lander2k2:
For the sake of UX with tptctl, add aliases that shorten the API object names so that users can type `tptctl get kri` instead of `tptctl get kubernetes-runtime-instances`.

Following are the abbreviations for API objects to use as aliases on the appropriate commands.

[tptctl-command-alias.md](https://github.com/user-attachments/files/29553463/tptctl-command-alias.md)

PR #463, opened 2026-07-01
feat: add short aliases to subcommands

sg40:


sg40:
https://github.com/threeport/threeport/issues/462

lander2k2 (changes requested):
@sg40 This looks great for all the `get` commands.  Now we need to do the same for all the `create`, `replace` and `delete` commands where applicable.  I wasn't specific about this in the issue.  Let's add constants at the top of each file and reference those to ensure consistency.  Take `aws.go` as an example...

```go
const (
    awsProviderShortAlias = "ap"
    awsEksKubernetesRuntimeShortAlias = "akr"
    awsEksKubernetesRuntimeDefinitionShortAlias = "akrd"
    awsEksKubernetesRuntimeInstanceShortAlias = "akri"
)
```

And then refer to these constants in each command that uses them for their alias.

PR #474, opened 2026-08-02
fix: move parent KRI cleanup from tptctl into GKE delete reconciliation

lander2k2:
Deleting a GcpGkeKubernetesRuntimeInstance previously relied on tptctl's config Delete method to block for up to 15 minutes waiting for the GKE instance to disappear, then confirm and delete the parent KubernetesRuntimeInstance. That only worked when the deletion originated from tptctl, and left the parent KRI orphaned when the GKE instance was deleted through any other path.

Move the cleanup into gkeLifecycle.OnDeleteConfirmed, which runs once the GCP cluster destroy actually completes. It looks up the parent KRI and triggers its deletion only when one is still present and not already scheduled for deletion; a not-found KRI or an in-progress deletion is treated as a no-op, since the normal KRI deletion flow removes the parent before the GKE destroy finishes.

Removes the corresponding wait-and-delete block from GcpGkeKubernetesRuntimeInstanceConfig.Delete, so tptctl returns as soon as the API accepts the delete rather than polling for it.

### DIFF
--- a/internal/gcp/gke_lifecycle.go
+++ b/internal/gcp/gke_lifecycle.go
@@ -1,6 +1,7 @@
 package gcp
 
 import (
+	"errors"
 	"fmt"
 	"time"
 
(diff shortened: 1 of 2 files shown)


randalljohnson (approved):
🚢

PR #477, opened 2026-08-04
fix: surface JetStream consumer failures instead of failing silently

sg40:


lander2k2 (changes requested):
Remove the ai directory.  Those requirements include information not relevant to this repo.

PR #479, opened 2026-08-12
fix: close api validation gaps and export provider mapping

randalljohnson:
Closes three gaps in the API tagged-field and validation layer, moves the cloud provider mapping tables somewhere a module can import them, and gives `tptctl` the GCP columns its data already carried.

`ProcessCoreTaggedFieldsBeforeUpdate()` did not run the persist-false hook, though the create path did. `SecretDefinition.Data` is the only field in the tree tagged `persist:"false"` and carries no `encrypt:"true"` tag, so that hook is the only thing keeping a secret payload out of the database. A `POST` carrying `Data` stored NULL and a `PATCH` carrying the same `Data` wrote the raw payload to the row. The update path now runs the hook.

`RedactEncryptedValues()` replaced each entry of a slice-shaped encrypted field with the placeholder `[encrypted value redacted]`, destroying the variable name along with the value. The only such fields are `MachineWorkloadDefinition.Env` and `MachineWorkloadInstance.Env`, whose entries are `KEY=VALUE`, so `tptctl get` printed a bare placeholder where `DB_PASSWORD=...` used to be. Feeding that back through `tptctl replace` then failed as `Env[0] is not in KEY=VALUE format`. Redaction now cuts at the first `=` and keeps the key, so the replay reaches `encryptValue()` and gets a 400 naming the entry. That parse error is now a bad-request error too, so the malformed input a caller can still send answers 400 rather than falling through to a 500.

`AwsProvider.beforeCreate()` rejects a provider carrying one of `AccessKeyID` and `SecretAccessKey` without the other. `beforeUpdate()` returned nil, so an update could leave half a credential in the row for the reconciler to pick up. It now re-enforces the same rule through `lib.IsFieldChanged()`, which detects the change under both the patch and the replace call shape; reading `tx.Statement.Changed()` directly would have covered the patch path only, since it always reports false under `Save`. Both hooks now call one shared check, which also drops about 40 lines of reflection from create.

`ValidLocation()` and the node-size table lived in `internal/kubernetes-runtime/mapping`, which Go's internal-package rule keeps out of reach of any threeport module living in its own repository. The package moves to `pkg/mapping/v0` unchanged apart from its package clause, and 12 call sites follow the import. The two Kubernetes runtime documentation pages follow it to the new path.

`tptctl config get-locations` and `get-machines` printed AWS and OCI columns while every entry already carried a GCP value, across 55 locations and 33 machine types. Both gain a GCP column and a GCP filter. Their more-than-one-filter check moves to cobra's `MarkFlagsMutuallyExclusive()`, because the hand-written one had drifted: `get-machines` listed the `get-locations` flags.

Tests cover each change: the persist-false wiring against a real `SecretDefinition` on both paths, the redaction round trip end to end, the access-key pairing rule across both call shapes, and the mapping validator. `pkg/api/lib/v0/persist_hooks_test.go` adds the first direct coverage of `ProcessPersistFalseTaggedFields()`, with a hooks-skipped negative control so the positive assertion is the hook's doing rather than something GORM would have done anyway.


randalljohnson:
@greptileai

PR #480, opened 2026-08-13
feat: add as-of-system-time pagination mode

randalljohnson:
Adds an as-of-system-time pagination strategy and makes it the default, so a paginated list stops copying the table it is reading.

Paging used to open each result set with two data definition language (DDL) statements: `CREATE MATERIALIZED VIEW <name> AS SELECT * FROM <table> ORDER BY ID ASC`, then a `CREATE INDEX` on it. Every paginated list request therefore copied the whole table, filtered or not, and left an object behind for a background sweeper to drop when a client stopped paging partway.

The events table is the largest in the system and the one that drove this. At 1,000,000 rows the `CREATE MATERIALIZED VIEW` plus its index costs 8.3 seconds before the first page is returned, against 52ms for the snapshot read, measured on CockroachDB v24.3.3. The gap tracks table size, since the view copies every row while the snapshot read touches only the page.

The new strategy captures a hybrid logical clock (HLC) timestamp with `cluster_logical_timestamp()` on the first page and reads every page `AS OF SYSTEM TIME` that value. That buys the same stable snapshot with no copy, no index build, and nothing to sweep. It costs snapshot lifetime: a view lives until its TTL, while a timestamp is readable only inside CockroachDB's garbage-collection window, so a client paging slowly through a large result set gets an expired session and has to restart. `-pagination-mode` keeps the view strategy available for that case.

Client errors get the same pass. An unparseable request body answers 400 rather than 500, and a bind error's response carries only the error message, where it used to append the wrapped cause and name Go struct and field types back to the caller. An expired pagination session answers 400 as well. A queryid the server can no longer resolve, whether its snapshot passed the garbage-collection window or its materialized view was dropped, tells the caller to restart pagination with no queryid rather than reporting a server fault.

Nothing downstream has to change. The mode is assigned to `Handler.PaginationMode` rather than passed to `handlers.New()`, so the constructor keeps the signature modules already compile against, and its zero value reads as as-of-system-time. Dropping the knob later does not move a signature either.


randalljohnson:
@greptileai

pkg/api-server/v0/handlers/handlers.go
```diff
@@ -151,3 +170,168 @@ func GenerateMaterializedViewName() (string, string) {
...
+// Under `PaginationModeMaterializedView` a view is created on the initial call
+// (empty pageParams.QueryId) and looked up by queryId on continuation; the view
+// is dropped once a page comes back shorter than the limit, so the TTL sweeper
+// doesn't have to, and a queryid naming no live view returns
+// `ErrPaginationSessionExpired`. Under `PaginationModeAsOfSystemTime` a fresh
+// HLC is captured on the initial call and the caller's queryId is validated and
+// reused on continuation; page-SELECT errors pass through
+// `TranslatePaginationSessionError()` so a GC-expired snapshot reaches the
+// client as the same expired-session error.
+func (h Handler) DispatchGetPaginatedRecords(
```
randalljohnson, line 288: Both modes built their own `SELECT *` and dropped the scoped query, so a filtered list went unfiltered past page one. This carries it through.

randalljohnson (commented):
Details

pkg/api-server/v0/handlers/handlers.go
```diff
@@ -94,37 +96,54 @@ func (h Handler) CreateMaterializedView(queryTable string) (string, string, erro
 }
 
 // GetMaterializedViewName finds the name of the materialized view created for a given pagination query ID.
+// The query ID comes from the client, so it is checked against the shape the server issues and then
+// matched as a bound parameter, keeping it out of the SQL text.
 func (h Handler) GetMaterializedViewName(queryId string) (string, error) {
```
randalljohnson, line 106: We used to interpolate the client's queryid into the `LIKE`, so a crafted id could close the literal and read `information_schema`. It's bound now, and we only accept ids we issued.

pkg/api-server/v0/handlers/handlers.go
```diff
@@ -9,24 +9,26 @@ import (
...
+	JS             nats.JetStreamContext
+	Logger         *zap.Logger
+	PaginationMode apiserver_lib.PaginationMode
 }
 
-// New returns a new Handler.
-func New(db *gorm.DB, nc *nats.Conn, rc nats.JetStreamContext, logger *zap.Logger) Handler {
-	return Handler{db, nc, rc, logger}
+// New returns a new Handler configured with the given pagination mode.
+func New(db *gorm.DB, nc *nats.Conn, rc nats.JetStreamContext, logger *zap.Logger, paginationMode apiserver_lib.PaginationMode) Handler {
```
randalljohnson: This is the breaking change for modules. The wrapper that calls it is generated, so a module picks it up on regen, not by hand.

pkg/sdk/v0/gen/pkg/api-server/handlers.go
```diff
@@ -2237,6 +2201,46 @@ const (
 	blockedDeleteCheckBackstop
 )
 
+// paginationErrorResponse returns the body of the error branch a paginated
+// list handler runs when the dispatcher fails. A queryid the client can
+// correct answers 400 without a log line, since it reports a bad request
+// rather than a fault: either the id names no snapshot the server issued, or
+// the snapshot it named has expired and the client has to start over with no
+// queryid. Everything else is a server fault, so it is logged and answers 500.
+func paginationErrorResponse(gen *gen.Generator) func(*Group) {
```
randalljohnson, line 2210: A bad or expired queryid is the client's to fix, so this is a 400 now, not a 500. Lands in all 18 generated handlers, and both hand-written callers match.

pkg/api-server/v0/handlers/events.go
```diff
@@ -267,50 +269,93 @@ func (h Handler) GetEventsJoinAttachedObjectReferences(c echo.Context) error {
...
 			}
 
-			// build and execute the CREATE MATERIALIZED VIEW. Raw SQL,
-			// so substitute the literal type into the shared join
-			// clause instead of binding via a gorm placeholder.
+			// build the join clause once; used by both mode branches
+			// below. Raw SQL, so substitute the literal event type in
+			// place of gorm's `?` placeholder.
 			joinClause := strings.Replace(
 				eventJoinAttachedObjectReferenceClause,
```
randalljohnson, line 276: Out of scope here. This path still drops the bound event filter, same defect the dispatcher fixes elsewhere. Already fixed on the events branch stacked on this one, PR opening soon.

PR #481, opened 2026-08-14
feat: harden sdk generators and module proxy

randalljohnson:
Hardens the SDK generators, the module reverse proxy, and the module-install test path.

Generated reconcilers skip an update when deletion is already scheduled, name the failed operation in their errors, and pass `authEnabled` into caller capture so an auth-off API stamps the control plane organizational unit and reconcilers can keep writing owned rows. `InitModuleRouter()` takes the same flag: with auth on, the proxy dials each module over https and presents the control plane client certificate; with auth off it stays on plain http.

Upgrading only the core API is a version floor for installed modules. The installer now exposes the module API Service on port 443 when auth is on, and `ModuleApi.Endpoint` stays a portless FQDN, so the proxy dials `https://<fqdn>` on 443. A module installed before this change still serves HTTP on port 80, and a register-only retry does not patch the Service. Reinstall every module after this lands.

An empty or non-PEM certificate authority now fails at startup in both the hand-written module proxy and the generated REST API server, instead of producing opaque TLS errors later.

`mage test:moduleGen` type-checks a generated module in Unit Tests. `mage test:moduleInstall` installs one at the end of Integration Tests, after `mage test:up` has written the Threeport config that both targets now read from disk. A miniature reconciler fixture under `internal/reconcilertest` compiles the same generators against a fake object so a skip-on-delete or `persist:"false"` emission break fails the unit suite.

Controller registration still looks a leftover `ModuleController` up by name globally and rebinds it when `ModuleApiID` differs. That repairs a partial install of the same module. A second module that chose the same controller name would take the first module's row.


randalljohnson:
@greptileai

randalljohnson:
@greptileai

randalljohnson:
@greptileai

lander2k2 (approved):

PR #486, opened 2026-09-08
fix: correct the database write path on cockroachdb

randalljohnson:
Adds a scoped uniqueIndex on Name so duplicate names fail at write rather than in a pre-read.

A pre-read uniqueness check races: two concurrent creates can both see no row and both persist. The unique index refuses the second write.

Object Name fields now carry `uniqueIndex:,where:deleted_at IS NULL`. A unique-index conflict answers 409 and names the API fields. A CockroachDB abort is retried. An update wakes its controller only when something changed.

That index is what CockroachDB lists as a unique constraint in `information_schema.table_constraints`. AutoMigrate's existing-table branch then emits `DROP CONSTRAINT uni_<table>_name`, and CockroachDB replies SQLSTATE 42704. The first pass on empty tables succeeds because that path is CreateTable and does not read the catalog. A second pass, including `goose up` after a partial install, fails with constraint does not exist. The same failure showed up on a module control-plane install: `goose up` against tables that already existed from a first pass. `TestRepeatedAutoMigrateIsRejected` in `test/cockroach/automigrate_test.go` pins that rejection.

`000001` therefore skips tables that exist and calls CreateTable for missing ones rather than AutoMigrate. Goose records version 1 only after `Up` returns nil, so this is the install and retry path, not a designed resume protocol. `AddMigrationNoTxContext` commits each CreateTable; `HasTable` skip plus CreateTable for missing tables is the CockroachDB-safe replacement. The TTL `ALTER TABLE` statements still run on every retry. `Up000001` uses the gorm handle from context, not goose's `db`. CreateTable includes the foreign key in the same `CREATE`, so the generated model list is sorted referenced-first and generate fails on a cycle. Later migrations must use SQL (`ALTER TABLE`, `CREATE INDEX`, `CREATE TABLE`) and never AutoMigrate.

GORM's CreateTable on a parent model does not add that model's many-to-many join tables. `createMissingTables` therefore runs two passes: parent tables first, then each join table. One pass fails foreign keys when the join's other parent has not been created yet. Schema-drift asserts those join tables exist, including `v0_module_api_routes_module_objects`.

Database state impact: `000001` is install and retry only. Goose will not re-run it on a database that already recorded version 1. In-place column adds on a kept-data database are a later numbered SQL file.

Wire format: a unique-index conflict answers 409 naming the API fields. The previous path returned 500. Response `Type` on GET and list is now the fully-qualified type name.

New tests cover a second AutoMigrate rejected on CockroachDB, CreateTable of only missing tables including join tables, `Up000001` run twice, unique-index 409 field mapping, uniqueness of ModuleApiRoute.Path across module APIs, and write retries on abort. CI runs the CockroachDB suite against a real server.


randalljohnson:
@greptileai

cmd/database-migrator/migrations/000001_init.go
```diff
@@ -12,29 +12,31 @@ import (
...
 	if err != nil {
 		return err
 	}
 
-	if err := gormDb.AutoMigrate(dbInterfaces000001()...); err != nil {
-		return fmt.Errorf("could not run gorm AutoMigrate: %w", err)
+	// create a table for each model that has none
+	for _, model := range dbInterfaces000001() {
+		if gormDb.Migrator().HasTable(model) {
+			continue
```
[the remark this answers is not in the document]
randalljohnson: Leave `000001` as install and retry only. Goose will not re-run version 1.

pkg/api/v0/common.go
```diff
@@ -50,3 +50,59 @@ type Reconciliation struct {
...
+}
+
+// ReconciliationUpdateNotifiable reports whether an update should notify the
+// controller. Spec edits notify even when markers are unchanged. A restamped
+// acknowledgement does not.
+func ReconciliationUpdateNotifiable(a, b Reconciliation) bool {
+	if ReconciliationStateChanged(a, b) {
+		return true
+	}
+	return !acknowledgementRefreshed(a, b)
```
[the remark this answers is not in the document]
randalljohnson: Compare reconciliation markers here only. The handler already requires `Reconciled` false, so a spec edit with no ack restamp still notifies.

.github/workflows/test.yml
```diff
@@ -66,6 +66,37 @@ jobs:
...
       run: go test -count=1 ./pkg/... ./internal/... ./cmd/...
 
+  cockroach-tests:
+    name: CockroachDB Tests
+    runs-on: ubuntu-24.04
+    env:
+      GOPATH: /home/runner/go
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v6.0.3
```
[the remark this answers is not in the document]
randalljohnson: Match the unit and integration jobs. Pin every third-party action to a SHA.

randalljohnson:
@greptileai

pkg/api/v0/common.go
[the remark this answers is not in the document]
randalljohnson: Leave the helper marker-only. Controllers send ack-only updates; users don't restamp ack.

pkg/sdk/v0/gen/pkg/api-server/handlers.go
randalljohnson: I dropped the pre-read name check. The partial unique index is what refuses a duplicate now.

sdk-config.yaml
```diff
@@ -116,10 +116,12 @@ ApiObjectGroups:
       Versions:
         - v0
       ExcludeRoute: true
+      ExcludeFromDb: true
```
randalljohnson: Definition and Instance are embed types, not tables. CreateTable would fail if they stayed in the init list.

pkg/sdk/v0/gen/generator.go
```diff
@@ -1090,7 +1096,66 @@ func (g *Generator) ValidateTags() error {
...
+
+const nameFieldName = "Name"
+
+// nameIndexTag is the gorm tag a Name field must carry: unique among undeleted rows.
+const nameIndexTag = "not null;uniqueIndex:,where:deleted_at IS NULL"
+
+const indexClassUnique = "UNIQUE"
+
+// nameIndexExemptions lists objects whose name uniqueness is not a table index.
+var nameIndexExemptions = map[string]string{
```
randalljohnson: ModuleObject stays unique per module API via its create hook, not a table index. I've noted a composite unique index for a follow-up PR.

pkg/sdk/v0/gen/generator.go
```diff
@@ -1090,7 +1096,66 @@ func (g *Generator) ValidateTags() error {
 			len(problems), strings.Join(problems, "\n  "),
 		)
 	}
-	return nil
+
+	return g.ValidateRelationshipCycles()
```
randalljohnson: CreateTable embeds the foreign key, so a cycle has to fail at generate time.

pkg/api-server/v0/handlers/kubernetes_workload.go
```diff
@@ -40,36 +41,48 @@ func (h Handler) AddKubernetesWorkloadResourceDefinitions(c echo.Context) error
...
 	}
 
 	// create all kubernetes workload resource definitions or none at all
 	var createdWRDs []v0.KubernetesWorkloadResourceDefinition
-	err := h.DB.Transaction(func(tx *gorm.DB) error {
-		for _, wrd := range k8sWorkloadResourceDefinitions {
-			if result := h.DB.Create(&wrd); result.Error != nil {
-				return result.Error
+	result := h.Write(c, func(db *gorm.DB) *gorm.DB {
+		createdWRDs = nil
```
randalljohnson: RetryWrite reruns this closure. I clear the slice so a retry doesn't append twice.

pkg/sdk/v0/gen/cmd/dbmigrator/migration.go
```diff
@@ -44,11 +44,15 @@ func GenDbMigratorMigration(gen *gen.Generator, sdkConfig *sdk.SdkConfig) error
...
 		Line(),
 
-		If(Err().Op(":=").Id("gormDb").Dot("AutoMigrate").Call(
-			Id(fmt.Sprintf("dbInterfaces%s", migrationVersion)).Call().Op("..."),
-		),
-			Err().Op("!=").Nil()).Block(
-			Return(Qual("fmt", "Errorf").Call(Lit("could not run gorm AutoMigrate: %w"), Err())),
+		Comment("create missing tables"),
+		If(
+			Err().Op(":=").Id("createMissingTables").Call(
```
randalljohnson: New initial migrations emit this. I updated 000001 by hand to match.

pkg/api-server/lib/v0/response.go
```diff
@@ -53,8 +53,8 @@ type Response struct {
 	// Meta contains PageRequestParams (current page and size of current page) and TotalCount (number of returned Object elements)
 	Meta Meta
 
-	// Type contains ObjectType of returned Data elements.
-	Type string `example:"KubernetesWorkloadInstance"`
+	// Type is <api-namespace>/<version>.<TypeName>, set even when Data is empty.
```
randalljohnson: Empty error bodies still need a type the client can parse.

pkg/api/v0/module.go
```diff
@@ -68,11 +68,12 @@ type ModuleController struct {
 
 // ModuleObject is an API object that is managed by a module in Threeport.  This provides
 // central registry of all API objects across all modules for each Threeport control plane.
+// The (Name, ModuleApiID) pair is unique.
 type ModuleObject struct {
 	Common `swaggerignore:"true" mapstructure:",squash"`
 
 	// The name of the API object.
-	Name *string `validate:"required" gorm:"not null"`
+	Name *string `validate:"required" gorm:"not null;uniqueIndex:idx_module_object_identity,where:deleted_at IS NULL"`
```
randalljohnson: Name uniqueness is (Name, ModuleApiID), same shape as ModuleApi. The create hook is gone.

pkg/api-server/lib/v0/unique_violation.go
```diff
@@ -0,0 +1,152 @@
...
+	"strings"
+	"sync"
+
+	"github.com/jackc/pgx/v5/pgconn"
+	"github.com/labstack/echo/v4"
+	"go.uber.org/zap"
+	"gorm.io/gorm/schema"
+)
+
+// Unique-index conflicts arrive as SQLSTATE 23505. The write path turns
```
randalljohnson: I map unique-index conflicts to 409, Name and non-Name alike. The handler pre-read is gone.

pkg/api/lib/v0/tags.go
```diff
@@ -10,6 +10,7 @@ const (
 	ValidateTag     TagKey = "validate"
 	QueryTag        TagKey = "query"
 	JsonTag         TagKey = "json"
+	GormTag         TagKey = "gorm"
```
randalljohnson: GormTag is what ValidateTags uses to find the Name uniqueIndex.

pkg/api/v0/module.go
```diff
@@ -68,20 +74,21 @@ type ModuleController struct {
 
 // ModuleObject is an API object that is managed by a module in Threeport.  This provides
 // central registry of all API objects across all modules for each Threeport control plane.
+// The (Name, Version, ModuleApiID) combination is unique.
 type ModuleObject struct {
 	Common `swaggerignore:"true" mapstructure:",squash"`
 
 	// The name of the API object.
-	Name *string `validate:"required" gorm:"not null"`
+	Name *string `validate:"required" gorm:"not null;uniqueIndex:idx_module_object_identity,where:deleted_at IS NULL"`
```
randalljohnson, line 84: Uniqueness is (`Name`, `Version`, `ModuleApiID`), so two versions of the same object can coexist.

pkg/api/v0/module_validate.go
```diff
@@ -162,28 +171,9 @@ func (m *ModuleController) beforeDelete(tx *gorm.DB) error {
 	return nil
 }
 
-// beforeCreate ensures no API object with the object name for a given
-// module API already exists before persisting an API object.
+// beforeCreate validates the ModuleObject before create.
 func (m *ModuleObject) beforeCreate(tx *gorm.DB) error {
```
randalljohnson: the Import example gofmt is incidental. I dropped the create hook; the unique index refuses duplicates now.

pkg/sdk/v0/config.go
```diff
@@ -157,10 +157,6 @@ type ApiObject struct {
 	// needs custom middleware enabled.
 	AllowCustomMiddleware *bool
```
randalljohnson, line 158: `AllowDuplicateModelNames` is gone. Name uniqueness is `uniqueIndex`.

pkg/sdk/v0/gen/cmd/dbmigrator/main.go
```diff
@@ -14,7 +14,11 @@ import (
 	"github.com/threeport/threeport/pkg/sdk/v0/util"
 )
 
-// GenDbMigratorMain generates source code for the DB migrator main package.
+// migrationTestPackage is imported by both generated schema-drift tests.
+const migrationTestPackage = "github.com/threeport/threeport/pkg/migrationtest/v0"
+
+// GenDbMigratorMain generates the database-migrator main package.
+// Core and modules both emit this as boilerplate.
 func GenDbMigratorMain(gen *gen.Generator, sdkConfig *sdk.SdkConfig) error {
```
randalljohnson: I generate the migrator main and the schema-drift test from here. Modules get the same files.

pkg/sdk/v0/gen/cmd/dbmigrator/migration.go
```diff
@@ -13,8 +13,8 @@ import (
 	"github.com/threeport/threeport/pkg/sdk/v0/util"
 )
 
-// GenDbMigratorMigration generates the migration used to set the database
-// schema before the API server starts.
+// GenDbMigratorMigration generates the scaffolding initial migration.
+// A new module emits this as 000001_init.go. An existing file is not overwritten.
 func GenDbMigratorMigration(gen *gen.Generator, sdkConfig *sdk.SdkConfig) error {
```
randalljohnson, line 18: Modules emit this as their initial migration. Core's `000001` matches.

pkg/sdk/v0/gen/generator.go
```diff
@@ -988,17 +987,7 @@ var allowedEmbed = map[string]bool{
...
-// Validates:
-//   - relationship: kind is recognized; modifier keys are recognized;
-//     the target type (`type:<TypeName>` modifier or the field name minus
-//     its "ID" suffix) names a registered API object; field type is *uint
-//   - encrypt: value matches EncryptTrue
-//   - validate: value matches a recognized validator value
-//   - persist: value matches PersistFalse (true is the default; omit the tag)
-//   - query: value matches queryNamePattern
+// ValidateTags reports every tag value the generator will not accept.
 func (g *Generator) ValidateTags() error {
```
randalljohnson: I dropped the per-tag inventory. The function reports every rejection; listing them here would rot.

cmd/database-migrator/migrations/000001_init.go
```diff
@@ -96,7 +90,6 @@ func dbInterfaces000001() []interface{} {
 		&v0.ControlPlaneDefinition{},
 		&v0.ControlPlaneInstance{},
 		&v0.ControlPlaneComponent{},
-		&v0.Definition{},
 		&v0.DomainNameDefinition{},
```
randalljohnson, line 93: `Definition` and `Instance` are embed types, not tables.

cmd/database-migrator/migrations/util_gen.go
```diff
@@ -23,3 +24,71 @@ func getGormDbFromContext(ctx context.Context) (*gorm.DB, error) {
 
 	return gormDb, nil
 }
+
+// createMissingTables creates a table for each model and many-to-many join that has none.
```
randalljohnson, line 28: Generate this for core and every module. Same helpers, not type-specific.

pkg/api/v0/module.go
```diff
@@ -35,12 +42,13 @@ type ModuleApi struct {
...
 
-// ModuleApiRoute represents a route supported by a module API.
+// ModuleApiRoute represents a route supported by a module API. Path is
+// unique among undeleted rows, across every module API.
 type ModuleApiRoute struct {
 	Common `swaggerignore:"true" mapstructure:",squash"`
 
 	// The URL path supported by the module API.
-	Path *string `validate:"required" gorm:"not null"`
+	Path *string `validate:"required" gorm:"not null;uniqueIndex:,where:deleted_at IS NULL"`
```
randalljohnson: Two concurrent creates can both see no row. The unique index refuses the second.

pkg/sdk/v0/gen/generator.go
```diff
@@ -1090,7 +1092,49 @@ func (g *Generator) ValidateTags() error {
...
+	fieldSchema, err := schema.Parse(reflect.New(fieldOnly).Interface(), &sync.Map{}, schema.NamingStrategy{})
+	if err != nil {
+		return []string{fmt.Sprintf(
+			"%s.%s: gorm rejected %s:%q: %v",
+			objectName, fieldName, lib.GormTag, gormTag, err,
+		)}
+	}
+
+	for _, index := range fieldSchema.ParseIndexes() {
+		if index.Class == indexClassUnique && index.Where != "" {
```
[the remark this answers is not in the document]
randalljohnson: Require `deleted_at IS NULL` in every unique-index `where`. Extra AND conditions stay valid.

.github/workflows/test.yml
```diff
@@ -30,10 +30,10 @@ jobs:
       GOPATH: /home/runner/go
     steps:
     - name: Checkout
-      uses: actions/checkout@v6.0.3
+      uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
```
randalljohnson: I pin every third-party action to a SHA. The comment names the tag.

.github/workflows/test.yml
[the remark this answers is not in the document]
randalljohnson: Pinned in this PR. See the SHA comments on each `uses` line.

pkg/api-server/v0/handlers/client_error_status_test.go
```diff
@@ -301,9 +272,9 @@ func TestHandlersAnswerClientErrorsWithClientStatus(t *testing.T) {
 	// PayloadCheck looks an object's tagged fields up by api version and object
 	// type, and answers 500 when that lookup misses, which would mask every
 	// write case below
-	registerTaggedFields(api_v0.ObjectTypeKubernetesWorkloadDefinition, new(api_v0.KubernetesWorkloadDefinition))
-	registerTaggedFields(api_v0.ObjectTypeKubernetesWorkloadResourceDefinition, new(api_v0.KubernetesWorkloadResourceDefinition))
-	registerTaggedFields(api_v0.ObjectTypeSecretDefinition, new(api_v0.SecretDefinition))
+	registerValidateTags(api_v0.ObjectTypeKubernetesWorkloadDefinition, new(api_v0.KubernetesWorkloadDefinition))
```
randalljohnson, line 275: Drop the local copy. `registerValidateTags` in `unique_violation_status_test.go` is the one.

pkg/sdk/v0/gen/generator.go
```diff
@@ -445,6 +449,19 @@ func (g *Generator) New(sdkConfig *sdk.SdkConfig) error {
 		}
 	}
 
+	/////////////// populate Generator.RelationshipDependencies ////////////////
```
randalljohnson: I parse foreign keys from the whole version dir here. CreateTable order uses this graph.

pkg/sdk/v0/gen/generator.go
```diff
@@ -580,12 +592,17 @@ func (g *Generator) New(sdkConfig *sdk.SdkConfig) error {
 				return apiObjects[i].TypeName < apiObjects[j].TypeName
 			})
 
-			// index this group's object names so a type in its own file still contributes tags
-			groupObjectNames := make(map[string]bool)
-			for _, c := range apiObjects {
-				groupObjectNames[c.TypeName] = true
+			// inspect source code
```
randalljohnson: I parse only the group file for tags. 0.7 walks every model file; this merge dropped that.

pkg/sdk/v0/gen/generator.go
```diff
@@ -1100,6 +1074,15 @@ func (g *Generator) ValidateTags() error {
 		}
 	}
 
+	// uniqueIndex on an embed is not in any group's StructTags
```
randalljohnson: Name lives on Definition. StructTags never see that uniqueIndex without this pass.

pkg/sdk/v0/gen/generator.go
```diff
@@ -1109,7 +1092,72 @@ func (g *Generator) ValidateTags() error {
...
 	}
-	return nil
+
+	return g.ValidateRelationshipCycles()
+}
+
+// undeletedRowPredicate is the unique-index conjunct that excludes soft-deleted rows.
+const undeletedRowPredicate = "deleted_at IS NULL"
+
+// nameIndexTag is a unique index among undeleted rows.
```
randalljohnson: Every Name field gets this tag. ValidateTags rejects anything weaker.

pkg/sdk/v0/gen/generator.go
```diff
@@ -1109,7 +1092,72 @@ func (g *Generator) ValidateTags() error {
...
+
+const indexClassUnique = "UNIQUE"
+
+// validateUniqueIndex reports a uniqueIndex tag that does not unique-index undeleted rows.
+func validateUniqueIndex(objectName, fieldName, gormTag string) []string {
+	if !strings.Contains(gormTag, uniqueIndexToken) {
+		return nil
+	}
+
+	// ask gorm what the tag builds; a string match would accept tags gorm ignores
```
randalljohnson: I let gorm parse the tag. A substring match would accept uniqueIndex gorm ignores.

pkg/sdk/v0/gen/generator.go
```diff
@@ -1109,7 +1092,72 @@ func (g *Generator) ValidateTags() error {
...
+	if foundUnique {
+		return nil
+	}
+
+	// unique among undeleted rows; anything else lets a soft-deleted value block reuse
+	return unscoped
+}
+
+// uniqueIndexScopesUndeleted reports whether a unique-index where clause includes undeleted rows.
+func uniqueIndexScopesUndeleted(where string) bool {
```
randalljohnson: deleted_at IS NULL can sit anywhere in the where. Extra AND clauses stay valid.

test/integration/serialization_test.go
```diff
@@ -1,115 +0,0 @@
-package main
```
randalljohnson, line 1: Lives at `test/cockroach/serialization_test.go`. Needs SERIALIZABLE Cockroach, not a live control plane.

pkg/sdk/v0/gen/generator.go
randalljohnson: I require deleted_at IS NULL in every unique-index where. Extra AND conditions stay valid.

cmd/database-migrator/migrations/000001_init.go
```diff
@@ -12,29 +12,26 @@ import (
...
 // init registers the migration with goose at startup.
 func init() {
 	goose.AddMigrationNoTxContext(Up000001, Down000001)
 }
 
-// Up000001 creates the initial database schema and sets row-level
-// time-to-lives for event rows and the attached object reference
-// rows that link events to their subjects.
+// Up000001 creates the initial schema and sets event row TTLs.
+// Existing tables are left alone so a partial run can finish.
```
randalljohnson, line 25: Goose will not re-run version 1. Leave existing tables so a partial run can finish.

pkg/api/v0/module_validate.go
```diff
@@ -52,28 +49,9 @@ func (m *ModuleApi) beforeDelete(tx *gorm.DB) error {
...
-	// no error returned from API route lookup - return conflict error
-	return util_v0.NewConflictError(
-		fmt.Sprintf(
-			"module API route already exists with path %s for module API with D %d",
-			*m.Path,
-			*m.ModuleApiID,
-		),
-	)
+	return nil
 }
```
randalljohnson, line 56: Drop the path pre-read. The unique index refuses duplicates now.

pkg/sdk/v0/gen/generator.go
randalljohnson, line 1: This file is three regions. FK graph, then tag parse of every model file in the version dir, then unique-index checks including `Name` on `Definition`.

pkg/sdk/v0/gen/generator.go
randalljohnson, line 453: Region 1 of 3: walk every `.go` in `pkg/api/<version>/` for `foreignKey` tags and build a graph of which type must exist before which. `CreateTable` later uses that order. Not unique-index validation.

pkg/sdk/v0/gen/generator.go
randalljohnson: Region 2 of 3: tag parse. I still open only `{group}.go`. `0.7` walked every model file; this merge dropped that.

pkg/sdk/v0/gen/generator.go
randalljohnson: Region 3 of 3, this PR: `Name` on `Definition` never hits `StructTags`. `schema.Parse` and `deleted_at IS NULL` follow.

pkg/sdk/v0/gen/generator.go
```diff
@@ -1100,6 +1117,15 @@ func (g *Generator) ValidateTags() error {
 		}
 	}
 
+	// uniqueIndex on an embed is not in any group's StructTags
```
randalljohnson, line 1120: Region 3 of 3: `Name` lives on the embedded `Definition` type, so it never appears in the group's `StructTags` map. Second pass over `EmbedTypes`, gorm `schema.Parse` for the tag, and `deleted_at IS NULL` on that unique index.

pkg/sdk/v0/gen/generator.go
```diff
@@ -641,29 +654,37 @@ func (g *Generator) New(sdkConfig *sdk.SdkConfig) error {
...
 						for _, spec := range genDecl.Specs {
 							switch spec.(type) {
+							// in the case we're looking at a struct type definition, inspect
 							case *ast.TypeSpec:
+								// if the spec is a type spec, get the type spec and
+								// its name
 								typeSpec := spec.(*ast.TypeSpec)
 								objectName = typeSpec.Name.Name
 
+								// skip types that belong to another group's file
```
randalljohnson, line 664: Region 2 of 3: open every hand-written `.go` in `pkg/api/<version>/` and collect gorm tags. Types in files other than `{group}.go` still get validated.

pkg/sdk/v0/gen/pkg/api-server/handlers.go
lander2k2: The generator template still hardcodes the short `objectType` for the `c.Bind()` failure path in every generated Create/Update/Replace handler (`pkg/sdk/v0/gen/pkg/api-server/handlers.go:708`, and two more sites), while every other response on the same endpoint was migrated to `fullyQualifiedType`. A malformed-JSON request is the one error path that still returns the old short type string, breaking the new wire contract the PR itself documents. One-line generator fix, propagates to ~53 generated files.

pkg/sdk/v0/gen/relationships.go
```diff
@@ -0,0 +1,312 @@
...
+
+			// has-many: the child's table holds the key
+			if child, ok := sliceElementModel(field.Type, modelNames); ok {
+				dependencies[child] = append(dependencies[child], typeName)
+				continue
+			}
+
+			// belongs-to: this table holds TypeNameID
+			if referenced, ok := singularModel(field.Type, modelNames); ok {
+				if keyFields[referenced+"ID"] {
```
lander2k2, line 208: `parseRelationshipDependencies` only records a dependency edge via a reciprocal slice/struct field; a bare `XxxID *uint` FK with no reciprocal field (e.g. `ModuleObject.ModuleControllerID` → `ModuleController`) is invisible to the graph. Today it's masked by the alphabetical tie-break happening to sort correctly; a future type pair without that luck would fail `Up000001`'s `CreateTable` with a real FK-ordering error, uncaught by any test.

pkg/api/v0/module_validate.go
lander2k2, line 96: This is not introduced by this PR - but this is a good time to address it:

`ModuleApiRoute.afterCreate` registers an in-process route as a non-transactional side effect of a GORM hook that runs pre-commit; the new retry loop just means it can fire more than once per request.

lander2k2 (changes requested):

pkg/sdk/v0/gen/pkg/api-server/handlers.go
randalljohnson: Fixed. Create, update, and replace bind errors now pass `fullyQualifiedType`. `PayloadCheck` and `ValidateBoundData` still take the short name.

pkg/sdk/v0/gen/relationships.go
randalljohnson: Fixed. A `TypeNameID` field whose `TypeName` is a model is now a dependency even with no reciprocal field.

pkg/api/v0/module_validate.go
randalljohnson: This one is a bit complicated, created an issue to track here https://github.com/threeport/threeport/issues/501.

randalljohnson:
@greptile-apps

randalljohnson:
@greptile-apps

pkg/sdk/v0/gen/relationships.go
```diff
@@ -0,0 +1,330 @@
...
+					keyFields[name.Name] = true
+				}
+			}
+		}
+
+		// bare TypeNameID with no reciprocal field still holds a foreign key
+		for keyField := range keyFields {
+			referenced := strings.TrimSuffix(keyField, "ID")
+			if referenced != "" && referenced != typeName && modelNames[referenced] {
+				appendUniqueDependency(dependencies, typeName, referenced)
```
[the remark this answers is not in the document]
randalljohnson: Dropped the ModelID-only loop. Table order now follows belongs-to and has-many associations only.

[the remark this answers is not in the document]
randalljohnson: v0 has 63 ModelID columns and 0 belongs-to pairs (*Model plus ModelID). CREATE order now follows the 38 has-many slices. A lone ModelID is an object-id column (relationship:requires), not a GORM foreign key.

lander2k2 (approved):

PR #487, opened 2026-09-09
test: harden infra provider lifecycle

randalljohnson:
Adds test seams and unit tests for the shared infrastructure create and delete state machines.

`HandleInfraCreate()` and `HandleInfraDelete()` keep their two-argument signatures, so the gcp and oci adapters compile unchanged. `LifecycleConfig` holds the stale-ack threshold, refresh interval, semaphore capacity, and persist retry budget. Production reads the current defaults. Tests swap the struct under a lock. Launch captures the semaphore channel once so a capacity swap cannot strand a slot. `NewPulumiWorkspace()` with `WithStateDirRoot()` points per-instance Pulumi state at a temp dir; an empty instance name returns an error instead of colliding on a shared state file.

The handlers stay two-argument because they are the provider-facing API, and only tests ever vary the tunables.

New tests cover every create and delete branch, semaphore backpressure and panic recovery, and a 2000-instance requeue loop whose peak in-flight count must equal the injected capacity.


internal/provider/pulumi_workspace.go
```diff
@@ -42,12 +43,54 @@ type PulumiWorkspace struct {
...
+
+// NewPulumiWorkspace returns a workspace for the named runtime instance
+// under the given pulumi project. The state file path resolves to
+// stateDir/.pulumi/stacks/<project>/<name>.json.
+//
+// The name is required and resolving a state dir without one returns an
+// error. The project is not enforced here, because the gke path builds a
+// workspace first and fills the project in afterwards; passing an empty
+// project yields a path with an empty project segment until it is set.
+func NewPulumiWorkspace(name, project string, opts ...PulumiWorkspaceOption) *PulumiWorkspace {
```
randalljohnson, line 68: I exported this for the GCE machine provider sitting on this branch. The gcp and oci adapters still construct PulumiWorkspace as a literal.

internal/provider/infra_lifecycle.go
```diff
@@ -406,7 +557,7 @@ func launchInfraDelete(config infraConfig) (int64, error) {
 
 	// launch deletion in background goroutine
 	go func() {
-		defer func() { <-infraSemaphore }()
+		defer func() { <-sem }()
 		defer func() {
 			if r := recover(); r != nil {
 				config.Log.Error(fmt.Errorf("panic: %v", r), "recovered panic in infrastructure delete goroutine")
```
randalljohnson, line 551: I skipped persistFailure() here on purpose. Delete has no persist callback, so recover only logs.

randalljohnson:
@greptileai

randalljohnson:
@greptileai

internal/provider/fakes_test.go
```diff
@@ -0,0 +1,668 @@
...
+// goroutines that stop on a channel close, so they exit promptly; the
+// window only covers scheduling.
+const goroutineDrainTimeout = 10 * time.Second
+
+// TestMain snapshots the goroutine count, runs the suite, then fails the
+// run when background goroutines outlive it. A leak here means a lifecycle
+// operation returned without stopping something it started, which is the
+// failure the semaphore and ack-refresh work is meant to rule out.
+func TestMain(m *testing.M) {
+	baselineGoroutines = runtime.NumGoroutine()
```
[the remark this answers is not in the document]
randalljohnson: I changed the leak check to match leftover lifecycle stacks, not the process-wide goroutine count.

lander2k2 (approved):

PR #502, opened 2026-09-15
feat: add gce machine infrastructure provider

randalljohnson:
Adds a Google Compute Engine backend for a machine runtime.

`GceMachineInfra` deploys a VM and an SSH firewall through Pulumi. SSH keys are generated in process: the public key is instance metadata, and the private key is returned from `CreateOutputs()` rather than as a stack output. `CreateGCPServiceAccountWithKey()` errors when the GCP service account already exists.


internal/provider/machine/machine_gce.go
```diff
@@ -0,0 +1,415 @@
+// Package machine provisions a Google Compute Engine VM for a Threeport
```
randalljohnson: 1. type and constructor (`GceMachineInfra`)
2. validate and deploy (`createInfra()`)
3. pulumi program (firewall + instance)
4. ssh keys
5. stack outputs

internal/provider/machine/machine_gce.go
```diff
@@ -0,0 +1,415 @@
...
+	_ provider.StreamableProvider  = (*GceMachineInfra)(nil)
+	_ provider.RefreshableProvider = (*GceMachineInfra)(nil)
+)
+
+// defaultSSHSourceRange is the firewall source CIDR when SSHSourceRanges is empty.
+const defaultSSHSourceRange = "0.0.0.0/0"
+
+// GceMachineInfra is the Google Compute Engine backend for a machine runtime.
+// It embeds PulumiWorkspace for stack, state, and automation API helpers.
+type GceMachineInfra struct {
```
randalljohnson: type and constructor: fields, `NewGceMachineInfra()`, stack config

internal/provider/machine/machine_gce.go
```diff
@@ -0,0 +1,415 @@
...
+// sshSourceRanges returns SSHSourceRanges, or the open CIDR when that slice is empty.
+func (i *GceMachineInfra) sshSourceRanges() []string {
+	if len(i.SSHSourceRanges) == 0 {
+		return []string{defaultSSHSourceRange}
+	}
+	return i.SSHSourceRanges
+}
+
+// validateRequiredFields reports every empty required field in one error.
+func (i *GceMachineInfra) validateRequiredFields() error {
```
randalljohnson: validate and deploy: required fields, `createInfra()`, destroy

internal/provider/machine/machine_gce.go
```diff
@@ -0,0 +1,415 @@
...
+// SetStackState restores stack state from JSON. It fills project defaults first.
+func (i *GceMachineInfra) SetStackState(state *datatypes.JSON) error {
+	// fill project defaults and stack config before writing state
+	i.ensurePulumiProjectDefaults()
+	i.syncStackConfigs()
+	return i.PulumiWorkspace.SetStackState(state)
+}
+
+// pulumiProgram defines the Pulumi resources for the GCE VM stack.
+func (i *GceMachineInfra) pulumiProgram() pulumi.RunFunc {
```
randalljohnson: pulumi program: SSH firewall `TargetTags` plus instance tags; private key stays off exports

internal/provider/machine/machine_gce.go
```diff
@@ -0,0 +1,415 @@
...
+			AccessConfigs().
+			Index(pulumi.Int(0)).
+			NatIp())
+
+		return nil
+	}
+}
+
+// generateSSHKeyPair returns a PKCS1 PEM private key and an authorized_keys public key.
+func generateSSHKeyPair() (privPEM, pubAuthorized string, err error) {
```
randalljohnson: ssh keys: generate, restore from PEM, derive public key

internal/provider/machine/machine_gce.go
```diff
@@ -0,0 +1,415 @@
...
+	}
+	pub, err := ssh.NewPublicKey(&key.PublicKey)
+	if err != nil {
+		return "", fmt.Errorf("failed to build SSH public key: %w", err)
+	}
+	return string(ssh.MarshalAuthorizedKey(pub)), nil
+}
+
+// captureOutputs copies hostname and externalIP from a Pulumi up result.
+func (i *GceMachineInfra) captureOutputs(outputs auto.OutputMap) {
```
randalljohnson: stack outputs: hostname and NAT IP only

internal/provider/machine/machine_gce_test.go
```diff
@@ -0,0 +1,654 @@
+package machine
```
randalljohnson: 1. pulumi mocks
2. program tests (firewall, metadata, no private key)
3. deploy validation and restore

internal/provider/machine/machine_gce_test.go
```diff
@@ -0,0 +1,654 @@
...
+	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
+	"golang.org/x/crypto/ssh"
+	"gorm.io/datatypes"
+
+	"github.com/threeport/threeport/internal/provider"
+)
+
+// requirePulumi skips the test when the pulumi CLI is not on PATH.
+// SetStackState shells out through the local workspace, so that path cannot run without it.
+func requirePulumi(t *testing.T) {
```
randalljohnson: mocks and `newTestInfra()`

internal/provider/machine/machine_gce_test.go
```diff
@@ -0,0 +1,654 @@
...
+	if seeded.sshPublicKeyAuthorized != "ssh-rsa AAAAseeded threeport" {
+		t.Fatal("ensureSSHKeyPair overwrote a pre-seeded public key")
+	}
+	if seeded.sshPrivateKeyPEM != "" {
+		t.Fatal("ensureSSHKeyPair generated a private key when the public key was already set")
+	}
+}
+
+// TestPulumiProgram_CreatesInstanceAndFirewall covers one instance and one SSH firewall from the program.
+func TestPulumiProgram_CreatesInstanceAndFirewall(t *testing.T) {
```
randalljohnson: program tests: instance, firewall tags, ssh metadata, no private-key export

internal/provider/machine/machine_gce_test.go
```diff
@@ -0,0 +1,654 @@
...
+	if len(firewalls) != 1 {
+		t.Fatalf("expected exactly 1 firewall, got %d", len(firewalls))
+	}
+	if got := firewallSourceRanges(t, firewalls[0]); !equalStringSlices(got, []string{"10.0.0.0/8"}) {
+		t.Errorf("firewall sourceRanges = %v, want [10.0.0.0/8]", got)
+	}
+}
+
+// TestDeployInfra_MissingRequiredFields rejects each required field when it is empty.
+func TestDeployInfra_MissingRequiredFields(t *testing.T) {
```
randalljohnson: deploy validation, restored PEM, region from zone

internal/provider/machine/machine_gce_scale_test.go
```diff
@@ -0,0 +1,355 @@
+package machine
```
randalljohnson: 1. 2000-name round trip
2. pulumi-backed round trip
3. concurrent read/write

internal/provider/machine/machine_gce_scale_test.go
```diff
@@ -0,0 +1,355 @@
...
+	if err := os.Rename(tmp, path); err != nil {
+		os.Remove(tmp)
+		return fmt.Errorf("failed to rename state: %w", err)
+	}
+	return nil
+}
+
+// TestStateDirIsolation_N2000_RoundTrip covers 2000 GCE machine workspaces
+// sharing one state dir root: each writes and reads only its own checkpoint.
+func TestStateDirIsolation_N2000_RoundTrip(t *testing.T) {
```
randalljohnson: 2000 isolated state dirs

internal/provider/machine/machine_gce_scale_test.go
```diff
@@ -0,0 +1,355 @@
...
+	if c := inFlight.Load(); c != 0 {
+		t.Fatalf("in-flight counter did not drain to 0: %d", c)
+	}
+
+	assertGoroutinesDrain(t, baselineGoroutines)
+}
+
+// TestStateDirIsolation_PulumiBacked_RoundTrip covers SetStackState then
+// ReadStateFile on many GCE workspaces sharing one state dir root.
+func TestStateDirIsolation_PulumiBacked_RoundTrip(t *testing.T) {
```
randalljohnson: pulumi-backed round trip

internal/provider/machine/machine_gce_scale_test.go
```diff
@@ -0,0 +1,355 @@
...
+	})
+	if failed {
+		t.FailNow()
+	}
+}
+
+// TestStateDirIsolation_ConcurrentReadWrite_Race covers a writer on one
+// instance and a reader on another sharing a root; the reader never sees
+// the writer's payload.
+func TestStateDirIsolation_ConcurrentReadWrite_Race(t *testing.T) {
```
randalljohnson: concurrent read/write race

randalljohnson:
@greptile-apps

internal/provider/gcp_bootstrap.go
```diff
@@ -458,17 +462,24 @@ func CreateGCPServiceAccountWithKey(projectID, accountName string) (*GCPServiceA
...
 	if err := grantServiceAccountRolesForProject(crmService, projectID, account.Email); err != nil {
 		return nil, fmt.Errorf("failed to grant IAM roles: %w", err)
 	}
 
-	// Create and export a key for the service account
+	if existed {
+		// prune user-managed keys so a new key stays under the 10-key limit
+		if err := pruneUserManagedServiceAccountKeys(iamService, projectID, account.Email); err != nil {
+			return nil, fmt.Errorf("failed to prune stale service account keys: %w", err)
+		}
```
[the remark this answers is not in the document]
randalljohnson: prune-all is the quota path. a unique-name 409 after this still mints a replacement key.

internal/provider/machine/machine_gce.go
```diff
@@ -0,0 +1,415 @@
...
+	i.StackConfigs = map[string]string{
+		"gcp:project": i.ProjectID,
+		"gcp:region":  region,
+	}
+}
+
+// sshSourceRanges returns SSHSourceRanges, or the open CIDR when that slice is empty.
+func (i *GceMachineInfra) sshSourceRanges() []string {
+	if len(i.SSHSourceRanges) == 0 {
+		return []string{defaultSSHSourceRange}
```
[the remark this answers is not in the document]
randalljohnson: open SSH is the default. set `SSHSourceRanges` to tighten it.

internal/provider/machine/machine_gce.go
```diff
@@ -0,0 +1,415 @@
...
+
+	// The ephemeral public IPv4 exported from the stack
+	externalIP string
+}
+
+// NewGceMachineInfra returns a GCE machine provider for the named runtime instance.
+func NewGceMachineInfra(name string, opts ...provider.PulumiWorkspaceOption) *GceMachineInfra {
+	return &GceMachineInfra{
+		PulumiWorkspace: *provider.NewPulumiWorkspace(name, "gce", opts...),
+	}
```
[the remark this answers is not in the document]
randalljohnson: this PR is the GCE backend only. runtime selection lands on the stacked machine-runtime branch.

internal/provider/machine/machine_gce.go
```diff
@@ -0,0 +1,415 @@
...
+		})
+		if err != nil {
+			return fmt.Errorf("failed to create GCP provider: %w", err)
+		}
+
+		// create SSH firewall rule scoped to this instance's network tag
+		sshTag := i.RuntimeInstanceName
+		sourceRanges := pulumi.ToStringArray(i.sshSourceRanges())
+		_, err = compute.NewFirewall(pctx, fmt.Sprintf("%s-ssh", i.RuntimeInstanceName), &compute.FirewallArgs{
+			Name:    pulumi.String(fmt.Sprintf("%s-ssh", i.RuntimeInstanceName)),
```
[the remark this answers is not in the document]
randalljohnson: GCE name rules stay on the caller. `RuntimeInstanceName` is the instance name and tag as-is.

internal/provider/gcp_bootstrap.go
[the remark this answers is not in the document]
randalljohnson: fixed: prune oldest USER_MANAGED keys until one slot is free. live keys under quota stay.

internal/provider/machine/machine_gce.go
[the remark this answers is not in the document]
randalljohnson: fixed: `SSHSourceRanges` is required. empty no longer defaults to `0.0.0.0/0`.

internal/provider/machine/machine_gce.go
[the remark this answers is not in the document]
randalljohnson: this PR is the GCE backend. `NewGceMachineInfra()` is the constructor; machine-runtime calls it on the stacked branch.

internal/provider/machine/machine_gce.go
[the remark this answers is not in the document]
randalljohnson: fixed: `RuntimeInstanceName` must match GCE RFC1035, 1 to 63 chars, lowercase.

randalljohnson:
@greptile-apps

randalljohnson:
Follow-up, not blocking: REST clients skip tptctl, so cloud uniqueness belongs on POST (handler or a later controller step), not only the CLI. Task filed for a written policy so this is not renegotiated per PR.

randalljohnson:
Race tests live in `*_race_test.go` with `//go:build race`. `mage test:race` walks those packages via `util.RunRaceTests()`. `threeport-sdk gen` emits the same target so modules can call it after they re-pin.

internal/provider/infra_lifecycle_reconciler_loop_test.go
```diff
@@ -245,52 +242,3 @@ func TestInfraLifecycleReconcilerLoop_2000CreateAndDelete_Mixed(t *testing.T) {
 	// wait for in-flight operations to drain
 	waitForInFlightZero(t, 10*time.Second)
 }
-
```
randalljohnson: moved `TestPerInstanceStateDirIsolation` to `pulumi_workspace_race_test.go` (`//go:build race`).

internal/provider/pulumi_workspace_race_test.go
```diff
@@ -0,0 +1,63 @@
...
+	"fmt"
+	"os"
+	"path/filepath"
+	"sync"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
```
randalljohnson, line 15: moved here from `infra_lifecycle_reconciler_loop_test.go` so `mage test:race` runs it.

magefiles/magefile_gen.go
```diff
@@ -32,6 +32,22 @@ type Dev mg.Namespace
...
+// Unit runs the unit tests across the threeport packages.
+func (Test) Unit() error {
+	if err := util.RunUnitTests(); err != nil {
+		return fmt.Errorf("failed to run unit tests: %w", err)
+	}
+	return nil
+}
+
+// Race runs go test -race on packages that contain *_race_test.go files.
+func (Test) Race() error {
```
randalljohnson, line 44: `*_race_test.go` is how this target finds packages. `//go:build race` is how `mage test:unit` skips the file. Go does not treat the filename as special; the tag applies to the whole file, so we need both.

pkg/api/v0/gcp.go
```diff
@@ -7,7 +7,7 @@ type GcpProvider struct {
 	Common `swaggerignore:"true" mapstructure:",squash"`
 
 	// The unique name of a GCP provider.
-	Name *string `validate:"required" gorm:"not null"`
+	Name *string `validate:"required" gorm:"not null;unique"`
```
randalljohnson: TODO: after fix-database-write-path merges first, switch this to `gorm:"not null;uniqueIndex:,where:deleted_at IS NULL"` so a soft-deleted name can be reused.

pkg/api/v0/gcp_validate.go
```diff
@@ -10,8 +10,18 @@ import (
 	util "github.com/threeport/threeport/pkg/util/v0"
 )
 
-// beforeCreate validates the GcpProvider before create.
+// beforeCreate rejects a duplicate GcpProvider name before persist.
 func (g *GcpProvider) beforeCreate(tx *gorm.DB) error {
```
randalljohnson: TODO: after fix-database-write-path merges first, the uniqueIndex on Name may make this count redundant. Revisit then.

internal/provider/gcp_bootstrap.go
```diff
@@ -458,17 +462,21 @@ func CreateGCPServiceAccountWithKey(projectID, accountName string) (*GCPServiceA
...
 
-	// Grant IAM roles to the service account
+	// grant IAM roles to the service account
 	if err := grantServiceAccountRolesForProject(crmService, projectID, account.Email); err != nil {
 		return nil, fmt.Errorf("failed to grant IAM roles: %w", err)
 	}
 
-	// Create and export a key for the service account
+	if existed {
+		return nil, fmt.Errorf("GCP service account %s already exists; pick a different provider name", account.Email)
```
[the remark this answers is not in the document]
randalljohnson: fixed: grant, key, or API-write failure deletes the service account this create added.

pkg/api/v0/gcp.go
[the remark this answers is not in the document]
randalljohnson: declined: uniqueIndex waits on #486. unique is AutoMigrate on a fresh install.

randalljohnson:
@greptile-apps

pkg/config/v0/gcp_provider.go
```diff
@@ -146,6 +146,19 @@ func (g *GcpProviderConfig) Create(
...
+				*gcpProviderValues.ProjectID,
+				*gcpProviderValues.Name,
+			); delErr != nil {
+				return nil, fmt.Errorf(
+					"failed to create gcp provider in threeport API: %w; failed to delete newly created service account: %v",
+					err,
+					delErr,
+				)
+			}
+		}
```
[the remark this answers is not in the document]
randalljohnson: declined GET-before-delete. labeled the instance; reuse an existing SA when its description records provisioned-by and threeport-name for this GcpProvider.

internal/provider/gcp_bootstrap.go
```diff
@@ -456,19 +460,35 @@ func CreateGCPServiceAccountWithKey(projectID, accountName string) (*GCPServiceA
...
 
-	account, err := createServiceAccountForProject(iamService, projectID, serviceAccountID, displayName, description)
+	account, existed, err := createServiceAccountForProject(iamService, projectID, serviceAccountID, displayName, description)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create service account: %w", err)
 	}
 
-	// Grant IAM roles to the service account
+	if existed && !serviceAccountOwnedBy(account, accountName) {
+		return nil, fmt.Errorf("GCP service account %s already exists; pick a different provider name", account.Email)
```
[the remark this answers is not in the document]
randalljohnson: fixed: reject a provider name that is not already unique lowercase [a-z0-9-].

randalljohnson:
@greptile-apps

internal/provider/gcp_bootstrap.go
```diff
@@ -732,6 +784,20 @@ func generateServiceAccountID(name string) string {
...
+	if name == "" {
+		return false
+	}
+	for _, r := range name {
+		if (r < 'a' || r > 'z') && (r < '0' || r > '9') && r != '-' {
+			return false
+		}
+	}
+	return generateServiceAccountID(name) == generateServiceAccountID(strings.ToLower(name))
+}
```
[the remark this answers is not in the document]
randalljohnson: fixed: reject a name whose service-account id would truncate.

randalljohnson:
@greptile-apps

randalljohnson:
@greptile-apps

randalljohnson:
@greptile-apps

randalljohnson:
@greptile-apps

internal/provider/machine/machine_gce.go
```diff
@@ -0,0 +1,425 @@
...
+// StreamableProvider and RefreshableProvider come from the embedded workspace.
+var (
+	_ provider.InfraProvider       = (*GceMachineInfra)(nil)
+	_ provider.StreamableProvider  = (*GceMachineInfra)(nil)
+	_ provider.RefreshableProvider = (*GceMachineInfra)(nil)
+)
+
+// gceNameRe is GCE's RFC1035 instance-name pattern, 1 to 59 characters.
+// The SSH firewall name is {name}-ssh and is also capped at 63.
+var gceNameRe = regexp.MustCompile(`^[a-z]([-a-z0-9]{0,57}[a-z0-9])?$`)
```
randalljohnson, line 40: Instance names cap at 59 so `{name}-ssh` stays within GCE firewall name length.

randalljohnson:
@greptile-apps

randalljohnson:
@greptile-apps

pkg/util/v0/race.go
```diff
@@ -0,0 +1,90 @@
...
+	"os"
+	"path/filepath"
+	"sort"
+	"strings"
+)
+
+const raceTestSuffix = "_race_test.go"
+
+// RaceTestPackages returns module-relative packages that contain a *_race_test.go file.
+func RaceTestPackages(root string) ([]string, error) {
```
randalljohnson, line 15: I keep race tests in `*_race_test.go` with `//go:build race` so `mage test:unit` skips them. `mage test:race` instruments only those packages. If we ran all unit tests with `-race` we would pay the extra compile-time penalty on all packages instead of just the ones we need it for.

internal/provider/gcp_bootstrap.go
```diff
@@ -115,7 +119,7 @@ func (i *KubernetesRuntimeInfraGKE) createGCPServiceAccount(iamService *iam.Serv
 	displayName := fmt.Sprintf(serviceAccountDisplayFormat, i.RuntimeInstanceName)
 	description := fmt.Sprintf("Service account for Threeport instance %s to manage GCP resources", i.RuntimeInstanceName)
 
-	account, err := createServiceAccountForProject(iamService, i.ProjectID, serviceAccountID, displayName, description)
+	account, _, err := createServiceAccountForProject(iamService, i.ProjectID, serviceAccountID, displayName, description)
```
lander2k2, line 122: This line is part of the diff: `createServiceAccountForProject`'s signature changed in this PR from `(*iam.ServiceAccount, error)` to `(*iam.ServiceAccount, bool, error)`, and every call site had to be updated to match. At this call site (the GKE cluster bootstrap path, `createGCPServiceAccount`), the PR explicitly discards the new `existed` bool with `_` and applies none of the three protections it wires up at the sibling call site in `CreateGCPServiceAccountWithKey` (tptctl path) just below:

- it never calls [`serviceAccountOwnedBy`](https://github.com/threeport/threeport/blob/6f9445f7674d600a84fa918a0d0c6f687428e650/internal/provider/gcp_bootstrap.go#L603-L610),
- it never validates the name with [`canonicalGCPAccountName`](https://github.com/threeport/threeport/blob/6f9445f7674d600a84fa918a0d0c6f687428e650/internal/provider/gcp_bootstrap.go#L757-L772) before deriving the service-account ID, and
- its description string ([line 120](https://github.com/threeport/threeport/blob/6f9445f7674d600a84fa918a0d0c6f687428e650/internal/provider/gcp_bootstrap.go#L120)) never embeds `GcpOwnershipDescription`.

**Failure scenario:** If a service account whose ID matches `formatServiceAccountID("threeport-svc-%s", RuntimeInstanceName)` already exists in the target GCP project — e.g. left over from a deleted/renamed cluster, or two runtime names that share the same 30-char-truncated prefix — `createGCPServiceAccount` silently adopts it and [`grantServiceAccountRoles`](https://github.com/threeport/threeport/blob/6f9445f7674d600a84fa918a0d0c6f687428e650/internal/provider/gcp_bootstrap.go#L131-L134) grants it the full `threeportServiceAccountRoles` list (this PR just added `compute.instanceAdmin.v1` and `compute.securityAdmin` to that list — [lines 39-42](https://github.com/threeport/threeport/blob/6f9445f7674d600a84fa918a0d0c6f687428e650/internal/provider/gcp_bootstrap.go#L39-L42)), handing broad project-wide control-plane privileges to an account Threeport never verified it owns.

This is exactly the confused-deputy risk the new ownership check closes for the tptctl path; the GKE path remains exposed. Note that a naive fix of just adding `if existed && !serviceAccountOwnedBy(...)` here would also need the description to start embedding `GcpOwnershipDescription`, otherwise every future idempotent redeploy would find its own previously-created account "unowned" and fail permanently.

internal/provider/machine/machine_gce.go
```diff
@@ -0,0 +1,425 @@
...
+	if err != nil {
+		return "", "", fmt.Errorf("failed to build SSH public key: %w", err)
+	}
+	pubAuthorizedBytes := ssh.MarshalAuthorizedKey(pub)
+
+	return string(privPEMBytes), string(pubAuthorizedBytes), nil
+}
+
+// ensureSSHKeyPair keeps a restored private key or generates a new pair.
+func (i *GceMachineInfra) ensureSSHKeyPair() error {
```
lander2k2, line 353: `ensureSSHKeyPair` only skips generation when `i.sshPrivateKeyPEM` is already non-empty on the in-memory struct, but nothing repopulates that field on a fresh process: [`SetStackState`](https://github.com/threeport/threeport/blob/6f9445f7674d600a84fa918a0d0c6f687428e650/internal/provider/machine/machine_gce.go#L240-L245) only restores Pulumi backend state (not this private field), and [`SetCreateOutputs`](https://github.com/threeport/threeport/blob/6f9445f7674d600a84fa918a0d0c6f687428e650/internal/provider/machine/machine_gce.go#L419-L425) — the only setter for `sshPrivateKeyPEM` — has no caller anywhere in the non-test codebase.

**Failure scenario:** On any retry that reconstructs a fresh `GceMachineInfra` after a partial failure or process restart (once this provider is wired into a reconciler), `ensureSSHKeyPair` generates a brand-new RSA keypair even though the instance was already created with the old public key in its metadata. Pulumi's `up` then pushes a real metadata-update API call, and any private key already handed to a caller via [`CreateOutputs()`](https://github.com/threeport/threeport/blob/6f9445f7674d600a84fa918a0d0c6f687428e650/internal/provider/machine/machine_gce.go#L413-L417) is silently invalidated with no signal that happened.

This is currently unreachable — `NewGceMachineInfra` has no production caller yet in this PR — so it's a design gap to close before the reconciler wiring lands rather than an active bug today.

internal/provider/gcp_bootstrap.go
```diff
@@ -579,12 +592,21 @@ func createServiceAccountForProject(
...
 
 	fmt.Printf("Created GCP service account: %s\n", account.Email)
 
-	return account, nil
+	return account, false, nil
+}
+
+// serviceAccountOwnedBy reports whether the account description records the
+// same ownership pair GcpResourceLabels would put on a Compute resource.
+func serviceAccountOwnedBy(account *iam.ServiceAccount, ownerName string) bool {
```
lander2k2, line 605: `serviceAccountOwnedBy` checks ownership via `strings.Contains(account.Description, GcpOwnershipDescription(ownerName))`, an unanchored substring match, and [`GcpOwnershipDescription`](https://github.com/threeport/threeport/blob/6f9445f7674d600a84fa918a0d0c6f687428e650/internal/provider/provider.go#L87-L95)'s format string ends with the raw name value with no trailing delimiter.

**Failure scenario:** `GcpOwnershipDescription("app")` = `"...threeport-name=app"` is a literal prefix of `GcpOwnershipDescription("app-prod")` = `"...threeport-name=app-prod"`, so `serviceAccountOwnedBy(accountOwnedByAppProd, "app")` incorrectly returns `true`. This only matters once two different owner names can land on the same account (e.g. via the ID-truncation collisions noted in Finding 1), but if that ever happens the ownership check silently passes instead of catching it.

lander2k2 (changes requested):

internal/provider/gcp_bootstrap.go
randalljohnson: fixed: GKE create now validates the runtime name, writes `GcpOwnershipDescription`, and refuses unowned reuse. leftover accounts without the suffix fail closed.

internal/provider/machine/machine_gce.go
randalljohnson: agreed: a reconstructed `GceMachineInfra` would mint a new key. the private key stays out of Pulumi state on purpose; it is a secret and belongs on an encrypted Threeport API field. this PR adds `SeedSSHKeyPair` and persist-before-up on the backend. a later branch stacked on this one writes encrypted `SSHKey` before Pulumi `up` and rehydrates it on `BuildInfra`, so that gap does not land on 0.7.

internal/provider/machine/machine_gce.go
```diff
@@ -0,0 +1,458 @@
...
+	}
+	i.sshPrivateKeyPEM = sshPrivateKeyPEM
+	i.sshPublicKeyAuthorized = string(ssh.MarshalAuthorizedKey(signer.PublicKey()))
+	return nil
+}
+
+// persistGeneratedSSHKey writes the in-memory private key when PersistSSHKey is set.
+func (i *GceMachineInfra) persistGeneratedSSHKey() error {
+	if i.PersistSSHKey == nil || i.sshPrivateKeyPEM == "" {
+		return nil
```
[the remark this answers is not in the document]
randalljohnson: agreed: `PersistSSHKey` is nil on this PR because there is no production caller yet. the private key stays out of Pulumi state on purpose; it is a secret and belongs on an encrypted Threeport API field. this PR adds `SeedSSHKeyPair` and persist-before-up on the backend. a later branch stacked on this one writes encrypted `SSHKey` before Pulumi `up` and rehydrates it on `BuildInfra`, so that gap does not land on 0.7.

internal/provider/gcp_bootstrap.go
randalljohnson: fixed: `serviceAccountOwnedBy` now matches `GcpOwnershipDescription` as a suffix, so `app` does not own `app-prod`. covered by `TestServiceAccountOwnedBy_RejectsNamePrefix`.

lander2k2 (approved):

